PR #76, opened 2018-06-12
@Devdhar - Refractory neuron should not be updated by input 

Hananel-Hazan:


djsaunde:
I disagree; there are two types of refractoriness (that I know of):

1. Refractory neurons do not spike, or
2. Refractory neurons do not integrate input.

The former is simpler and IMO the better interpretation for our needs.

Alternatively, we can implement both forms of refractoriness.

djsaunde:
Moreover, there's a typo, which causes the build to fail. Did you run the tests before pushing?

bindsnet/network/nodes.py
```diff
@@ -188,7 +188,7 @@ def step(self, inpts, dt):
         self.v.masked_fill_(self.s, self.reset)
 
         # Integrate input and decay voltages.
-        self.v += inpts
+        self.v = self.v.masked_fill(self.refrac_count  ==0, self.v + inputs)
```
djsaunde: `masked_fill` can't be used here because it expects its second argument to be a single `float`, not a `torch.Tensor`.

bindsnet/network/nodes.py
```diff
@@ -238,28 +238,29 @@ def step(self, inpts, dt):
...
 
         # Refractoriness and voltage reset.
+
         self.refrac_count.masked_fill_(self.s, self.refrac)
         self.v.masked_fill_(self.s, self.reset)
-        
+
         # Integrate inputs.
-        self.v += inpts
+        self.v = self.v.masked_fill(self.refrac_count  ==0, self.v + inputs)
```
djsaunde: Same here.

bindsnet/network/nodes.py
```diff
@@ -404,12 +405,12 @@ def step(self, inpts, dt):
         self.refrac_count.masked_fill_(self.s, self.refrac)
         self.v.masked_fill_(self.s, self.reset)
         self.theta += self.theta_plus * self.s.float()
-        
+
         # Integrate inputs.
-        self.v += inpts
+        self.v = self.v.masked_fill(self.refrac_count  ==0, self.v + inputs)
```
djsaunde: And here.

bindsnet/network/nodes.py
```diff
@@ -479,19 +480,19 @@ def step(self, inpts, dt):
...
         if torch.sum(self.s) > 0:
             s = torch.zeros(self.s.size())
             s = s.view(-1)
             s[torch.multinomial(self.s.float().view(-1), 1)] = 1
             self.s = s.view(self.s.size()).byte()
-        
+
         # Integrate inputs.
-        self.v += inpts
+        self.v = self.v.masked_fill(self.refrac_count  ==0, self.v + inputs)
```
djsaunde: And here.

PR #77, opened 2018-06-14
Fix bugs and code improvements in Izhikevich neuron model

Hananel-Hazan:


bindsnet/network/nodes.py
```diff
@@ -181,14 +181,14 @@ def step(self, inpts, dt):
         self.refrac_count[self.refrac_count != 0] -= dt
 
         # Check for spiking neurons.
-        self.s = (self.v >= self.thresh) & (self.refrac_count == 0)
```
djsaunde: This `& (self.refrac_count == 0)` is needed. It's not a bug. Neurons shouldn't spike if they're refractory.

bindsnet/network/nodes.py
```diff
@@ -181,14 +181,14 @@ def step(self, inpts, dt):
...
+        self.s = (self.v >= self.thresh)
 
         # Refractoriness and voltage reset.
         self.refrac_count.masked_fill_(self.s, self.refrac)
         self.v.masked_fill_(self.s, self.reset)
 
         # Integrate input and decay voltages.
-        self.v += inpts
+        self.v += (self.refrac_count  == 0).float() * inpts
```
djsaunde: There's an extra space here: `(self.refrac_count  == 0).float() *` should be `(self.refrac_count == 0).float() *`

bindsnet/network/nodes.py
```diff
@@ -249,7 +249,7 @@ def step(self, inpts, dt):
         self.refrac_count[self.refrac_count != 0] -= dt
 
         # Check for spiking neurons.
-        self.s = (self.v >= self.thresh) & (self.refrac_count == 0)
```
djsaunde: Same here.

bindsnet/network/nodes.py
```diff
@@ -398,7 +398,7 @@ def step(self, inpts, dt):
         self.refrac_count[self.refrac_count != 0] -= dt
 
         # Check for spiking neurons.
-        self.s = (self.v >= self.thresh + self.theta) & (self.refrac_count == 0)
```
djsaunde: Here.

bindsnet/network/nodes.py
```diff
@@ -473,7 +473,7 @@ def step(self, inpts, dt):
...
```
djsaunde: And here.

bindsnet/network/nodes.py
```diff
@@ -525,11 +525,11 @@ def __init__(self, n=None, shape=None, traces=False, excitatory=True, rest=-65.0
         '''
         super().__init__(n, shape, traces, trace_tc)
 
-        self.rest = rest       # Rest voltage.
+        self.rest = rest         # Rest voltage.
```
djsaunde: Can you align these comments?

bindsnet/network/nodes.py
```diff
@@ -557,18 +556,18 @@ def step(self, inpts, dt):
             | :code:`inpts` (:code:`torch.Tensor`): Inputs to the layer.
             | :code:`dt` (:code:`float`): Simulation time step.
         '''
-        # Decrement refrac counters.
-        self.refrac_count[self.refrac_count != 0] -= dt
 
         # Check for spiking neurons.
-        self.s = (self.v >= self.thresh) & (self.refrac_count == 0)
```
djsaunde: Here.

bindsnet/network/nodes.py
```diff
@@ -557,18 +556,18 @@ def step(self, inpts, dt):
...
-        self.refrac_count.masked_fill_(self.s, self.refrac)
-        self.v.masked_fill_(self.s, self.reset)
+        self.v = torch.where(self.s, self.c, self.v)
+        self.u = torch.where(self.s, self.u + self.d, self.u)
+
 
         # Apply v and u updates.
-        self.v += dt * (0.04 * (self.v ** 2) + 5 * self.v + 140 - self.u + inpts)
+        self.v += dt * 0.5 * (0.04 * (self.v ** 2) + 5 * self.v + 140 - self.u + inpts)
+        self.v += dt * 0.5 * (0.04 * (self.v ** 2) + 5 * self.v + 140 - self.u + inpts)
```
djsaunde: This should only happen once.

bindsnet/network/nodes.py
djsaunde: Actually, I see why this works now. Disregard these comments!

bindsnet/network/nodes.py
Hananel-Hazan: I'm sorry, I dont see the different between the two snippets and I dont see the extra space

djsaunde: In your code, there's an extra space between `self.refrac_count` and the `==` operator.

bindsnet/network/nodes.py
Hananel-Hazan: Our default dt = 1 ms, we getting the same computational instability that Izhikavitch has in the Matlab code

djsaunde: I anticipated this; that's why you should use `dt = 0.5`, instead of running the statement twice for all `dt`.

Hananel-Hazan: dt can only be integers.. otherwise we getting bugs...

djsaunde: No, `dt` is a `float`, or it should be, at least.

djsaunde: What bugs?

PR #87, opened 2018-06-22
Renaming to locally connected layer and adding reference to reflect literature.

djsaunde:
Renaming method to reflect the layer type in the neural networks literature.


Hananel-Hazan:
OK

PR #90, opened 2018-06-24
Ngram, still a bug

hqkhan:
There's a bug still that I haven't been able to find. It breaks around 4250th image.


examples/mnist/eth_ngram_mnist.py
```diff
@@ -0,0 +1,171 @@
...
+        plt.pause(1e-8)
+
+    network._reset()  # Reset state variables.
+
+print('Progress: %d / %d (%.4f seconds)\n' % (n_train, n_train, t() - start))
+print('Training complete.\n')
+
+print ('Saving necessary items.\n')
+
+with open('/mnt/nfs/work1/rkozma/hqkhan/output.txt', 'w') as out:
```
djsaunde: Is there any reason you're saving things here, rather than re-using the results-saving code in the `bindsnet_experiments` repo?

bindsnet/evaluation/__init__.py
```diff
@@ -1,12 +1,13 @@
+import sys
```
djsaunde: Unused?

examples/mnist/eth_ngram_mnist.py
```diff
@@ -0,0 +1,171 @@
+import os
+import sys
```
djsaunde: Unused?

PR #98, opened 2018-07-19
Changing comment blocks to triple-quoted (as per PEP 8).

djsaunde:


hqkhan:
Cool. Will keep that in mind.

PR #112, opened 2018-08-16
Numenta Encoder

sharath:
Implementation of Numenta's geospatial encoder

The pyproj dependency on PyPI is broken, so I'm using the one on Github until it's fixed.


sharath:
reconsidering structure

PR #117, opened 2018-08-22
Add Izhikevich Metabolic Nodes, clamp and fixed examples

Hananel-Hazan:
1. Add Izhikevich Metabolic Nodes and remove unnecessary parameters in Izhikevich and IzhikevichMetabolic
2. Change the behavioral of clamp, now clamp is forcing certain voltage value in specific neuron at certain time. 
3. Fix examples in Network to reflect the last changes in BindsNET improvements


bindsnet/network/__init__.py
```diff
@@ -241,17 +245,19 @@ def run(self, inpts: Dict[str, torch.Tensor], time: int, **kwargs) -> None:
...
         # Simulate network activity for `time` timesteps.
         for t in range(timesteps):
-            # Update each layer of nodes.
+
+            # iterate on each layer
             for l in self.layers:
+                # Force neurons to certain voltage.
+                clamp = clamps.get(l, None)
+                if clamp is not None:
+                    self.layers[l].v = torch.where(clamp[t,:], clamps_v[l][t,:], self.layers[l].v)
```
djsaunde: Should be `self.layers[l].v = torch.where(clamp, clamps_v[l], self.layers[l].v`; the implementation of `clamps` (thus far) assumes there is no time dimension.

bindsnet/network/nodes.py
```diff
@@ -703,3 +697,80 @@ def reset_(self) -> None:
...
+            self.c = -65.0 + 15 * (self.r ** 2)
+            self.d = 8 - 6 * (self.r ** 2)
+        else:
+            self.r = torch.rand(n)
+            self.a = 0.02 + 0.08 * self.r
+            self.b = 0.25 - 0.05 * self.r
+            self.c = -65.0 * torch.ones(n)
+            self.d = 2 * torch.ones(n)
+
+        m = 0.7
```
djsaunde: Should be `self.m = 0.7`; otherwise, it can't be accessed elsewhere.

bindsnet/network/__init__.py
Hananel-Hazan: That right, now clamp have time dimension

bindsnet/network/nodes.py
Hananel-Hazan: You right

bindsnet/network/__init__.py
djsaunde: Why? The input is passed in, timestep by timestep.

Also, this breaks some of the examples.

PR #120, opened 2018-08-29
Plot non-square filters; fixing code style; fixing plotting bug.

djsaunde:


djsaunde:
Any comments on this? The biggest change is to the `bindsnet.learning`, where methods have been replaced by classes to avoid some control flow statements (for optimization purposes).

PR #125, opened 2018-09-06
Adding LocallyConnectedConnection and two-layer LocallyConnectedNetwork model.

djsaunde:


bindsnet/models/__init__.py
```diff
@@ -110,29 +111,26 @@ def __init__(self, n_inpt: int, n_neurons: int = 100, exc: float = 22.5, inh: fl
...
         """
         super().__init__(dt=dt)
 
         self.n_inpt = n_inpt
         self.n_neurons = n_neurons
-        self.n_CANs = n_CANs
-        self.running_time = running_time
+        self.n_cans = n_cans
 
         self.add_layer(Input(n=self.n_inpt), name='X')
```
Hananel-Hazan: `self.add_layer(Input(n=self.n_inpt), name='X')`
should be `RealInput` insted of `Input` 
But its not running... discussion is needed

bindsnet/models/__init__.py
```diff
@@ -1,10 +1,11 @@
...
 import torch
+import numpy as np
+
+from typing import Optional, Union, Tuple
 
 from ..network import Network
 from ..learning import PostPre
-from ..network.topology import Connection
+from ..network.topology import Connection, LocallyConnectedConnection
 from ..network.nodes import Input, LIFNodes, DiehlAndCookNodes, IzhikevichNodes
```
Hananel-Hazan: Need to add `RealInput`

Issue #130, opened 2018-09-17
Easier way to control learning in networks


sharath:
Right now you have to iterate through each connection's update rule and set it to NoOp, which is tedious. If you want to turn on learning, you have to restore the learning rule or reload it from disk.

sharath:
Added in #132

PR #132, opened 2018-09-17
Option to turn off learning in network

sharath:
#130

### DIFF
--- a/bindsnet/network/topology.py
+++ b/bindsnet/network/topology.py
@@ -78,11 +78,8 @@ class AbstractConnection(ABC):
         """
         learning = kwargs.get('learning', True)
         reward = kwargs.get('reward', None)
-        if learning or 'bindsnet.learning.NoOp' in str(self.update_rule.__class__):
+        if learning:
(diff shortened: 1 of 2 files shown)


bindsnet/network/topology.py
```diff
@@ -76,8 +76,13 @@ def update(self, **kwargs) -> None:
         """
         Compute connection's update rule.
         """
+        learning = kwargs.get('learning', True)
         reward = kwargs.get('reward', None)
-        self.update_rule.update(reward=reward)
+        if learning or 'bindsnet.learning.NoOp' in str(self.update_rule.__class__):
```
djsaunde: I think we should just have:

```
if learning:
    self.update_rule.update(reward=reward)
```

and just stipulate that if learning if "off", weight decay is off as well. In that sense, `learning = False` implies that weights are frozen.

sharath: Alright, but doesn't that make `NoOp` redundant? Why not just have `update_rule` be None then?

djsaunde:
Sounds good to me. Merging now.

Issue #140, opened 2018-10-10
Implementation of MSTDPET is wrong?


godelicbach:
First of all, thank you for making this open-source library. I really appreciate the intention and approach of this project. And I hope I could contribute something.

Maybe this problem would be more proper to be handled by pull request or something. But I'm not really used to github's contributing process or something. Please tell me if I'm doing something wrong.

Here's the thing. I looked inside the MSTDPET learning rule and I found its implementation is different from Florian's original learning rule. Current implementation is as follows. 
`        self.p_plus = -(self.tc_plus * self.p_plus) + a_plus * source_x`
`        self.p_minus = -(self.tc_minus * self.p_minus) + a_minus * target_x`

However, in Florian's paper, it's described as 
![image](https://user-images.githubusercontent.com/24662250/46744984-25e3d500-cce7-11e8-8dbb-d5e3ba5e5ad1.png)
or 
![image](https://user-images.githubusercontent.com/24662250/46745054-490e8480-cce7-11e8-87b7-8edb6b93b8b3.png)

Maybe someone who implemented it was confused between dP/dt and just P.

Thank you for reading!

djsaunde:
Hi @godelicbach,

Thanks for the catch! I think this is the best way to handle this problem, by the way, since it makes it easy for me to understand the issue before rewriting any code.

The implementation does seem wrong, like you say; I'll open a pull request with a fix and reference this issue in it, so you can check it out. I think one reason why it might be wrong is that we haven't really been using it in our lab for experiments. I simply implemented it one day because I read the M-STDP paper and thought it was an interesting idea.

godelicbach:
Thank you for fast and kind reply!
I also have found other minor bugs here and there. I'll make pull request when my branch's got stabilized.

djsaunde:
I'll just leave off with this: the `Pipeline` object and interface with OpenAI gym is not a focus of our development at the moment, so I expect there will be a lot of bugs (if this is what you are interested in using). I would love any and all pull requests.

PR #141, opened 2018-10-10
M-STDP-ET fix + code cleanup.

djsaunde:
Fix for #140.


djsaunde:
@sharath This might break some of your code. I changed file names ending with '.p' to '.pt', which is the PyTorch convention, since they implement their own custom serialization, to differentiate from pickle files.

PR #146, opened 2018-10-26
Breakout RL experiments

dee0512:
Major changes: 

1. Make the softmax decision on "sum of spikes" using spike monitors. 
2. Add 2 new breakout experiments.


bindsnet/pipeline/__init__.py
```diff
@@ -75,17 +75,20 @@ def __init__(self, network: Network, environment: Environment, encoding: Callabl
         else:
             self.history = {}
 
+        # monitor spikes for selecting action based on sum of spikes
```
djsaunde: It seems like this shouldn't be done in the `Pipeline` object, but outside in an experimental script. If the layers of the network are really large, for example, this could make things really slow.

examples/breakout/breakout_stdp_right.py
```diff
@@ -0,0 +1,56 @@
+import torch
```
djsaunde: What's the purpose of this script? We want to have only a few examples in the BindsNET repo so we don't have to maintain a bunch.

bindsnet/pipeline/__init__.py
dee0512: Yes. You are right. I will make the change.

examples/breakout/breakout_stdp_right.py
dee0512: @Hananel-Hazan wanted an experiment for reinforcing only the right action via STDP. This was to see if it performs well enough to include the results in the paper. At this point, I am not really sure which one we should use. The breakout.py example performs the best. Which is not saying much because it has no learning.

djsaunde: Got it, thanks. It might be better to keep this script in the [bindsnet_experiments repo](https://github.com/djsaunde/bindsnet_experiments).

dee0512: Ok. I will remove it for now. Since I am not sure about the utility of this experiment.

Hananel-Hazan: It should be in directory with the rest of the benchmark scripts and example scripts that will appear in the paper for Frontier. @djsaunde will create a spacial tag for it

Issue #156, opened 2018-11-21
Consistency with nu and nu_post, nu_pre


sharath:
There are some inconsistencies with nu_post and nu_pre. For instance, the Network is implemented using a tuple (nu_pre, nu_post) whereas some of the models use nu_pre, nu_post as separate parameters. (Also since they have kwargs it silently does nothing if one is used instead of the other)

I think we should try to keep it consistent.

Unless this is intentional?

djsaunde:
Yeah, I think this makes sense. The reason for the inconsistency is because both used to use `nu_pre, nu_post`, and I recently switched the `topology` modules and `LearningRule`s (I think this is what you meant by `Network`?) use `nu`. I simply hadn't switched them yet, since the `models` are pretty low priority / are there for our own experimental convenience.

PR #157, opened 2018-11-21
Consistency changes for Learning Rate

sharath:
Proposed changes for #156

Removed some dask temp files


examples/asteroids/random_network_baseline.py
```diff
@@ -56,7 +56,7 @@
 
 # Excitatory -> readout.
 exc_readout_conn = Connection(source=layers['E'], target=layers['R'], w=0.01 * torch.rand(layers['E'].n, layers['R'].n),
-                              update_rule=Hebbian, nu_pre=1e-2, nu_post=1e-2, norm=0.5 * layers['E'].n)
+                              update_rule=Hebbian, nu=(1e-2, 1e-2), norm=0.5 * layers['E'].n)
```
djsaunde: In general, you should use lists, not tuples, so they can be modified at training time.

PR #165, opened 2018-11-28
A few small fixes / tweaks.

djsaunde:


djsaunde:
@sharath not sure what you mean. However, we could have `PoissonInput` / `BernoulliInput` / `RankOrderInput` / etc., which would subclass both `Nodes` and `AbstractInput`. Is this what you mean? I'm not sure where `RealInput` factors in here.

PR #171, opened 2018-12-31
FIX: import for mnist reservoir example 

kingjr:


Hananel-Hazan:
Thanks @kingjr for the PR

PR #182, opened 2019-01-29
Hananel

Hananel-Hazan:


bindsnet/network/__init__.py
```diff
@@ -167,6 +167,28 @@ def save(self, file_name: str) -> None:
         """
         torch.save(self, open(file_name, 'wb'))
 
+    def load(file_name: str) -> None:
```
djsaunde: This function doesn't do anything. You're using `torch.load()` and then discarding the result.

djsaunde: You can remove this function and just use `bindsnet.network.load_network` instead.

bindsnet/network/__init__.py
```diff
@@ -1,6 +1,6 @@
 import torch
-
 from typing import Dict
+import tempfile
```
djsaunde: Move this up to immediately after `import torch`

bindsnet/network/topology.py
```diff
@@ -49,6 +56,9 @@ def __init__(self, source: Nodes, target: Nodes, nu: Optional[Union[float, Seque
         self.wmax = kwargs.get('wmax', None)
         self.norm = kwargs.get('norm', None)
         self.decay = kwargs.get('decay', None)
+        self.norm_by_max = kwargs.get('norm_by_max', None)
```
djsaunde: Should be `self.norm_by_max = kwargs.get('norm_by_max', False)`, not `None`.

bindsnet/network/topology.py
```diff
@@ -49,6 +56,9 @@ def __init__(self, source: Nodes, target: Nodes, nu: Optional[Union[float, Seque
...
+        self.norm_by_max_from_shadow_weights = kwargs.get('norm_by_max_from_shadow_weights', None)
```
djsaunde: Same here.

bindsnet/network/topology.py
```diff
@@ -141,6 +158,10 @@ def __init__(self, source: Nodes, target: Nodes, nu: Optional[Union[float, Seque
 
         self.b = kwargs.get('b', torch.zeros(target.n))
 
+        if self.norm_by_max_from_shadow_weights is not None:
```
djsaunde: Change to `if self.norm_by_max_from_shadow_weights:`

PR #183, opened 2019-02-07
Add node type argument to network connection

dee0512:
You can now have any type of neuron in network conversion (not just subtractiveIF)

### DIFF
--- a/bindsnet/conversion/__init__.py
+++ b/bindsnet/conversion/__init__.py
@@ -398,10 +398,7 @@ def _ann_to_snn_helper(prev, current, node_type, **kwargs):
     :return: Spiking neural network layer and connection corresponding to ``prev`` and ``current`` PyTorch modules.
     """
     if isinstance(current, nn.Linear):
-        if node_type:
-            layer = node_type(n=current.out_features, reset=0, thresh=1, refrac=0, **kwargs)
-        else:
-            layer = SubtractiveResetIFNodes(n=current.out_features, reset=0, thresh=1, refrac=0)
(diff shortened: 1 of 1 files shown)


bindsnet/conversion/__init__.py
```diff
@@ -388,7 +388,7 @@ def set_requires_grad(module, value):
     return ann
 
 
-def _ann_to_snn_helper(prev, current):
+def _ann_to_snn_helper(prev, current, node_type, **kwargs):
```
djsaunde: How about `node_type=`SubtractiveResetIFNodes` as default?

bindsnet/conversion/__init__.py
```diff
@@ -411,9 +414,12 @@ def _ann_to_snn_helper(prev, current):
         height = (input_width - current.kernel_size[1] + 2 * current.padding[1]) / current.stride[1] + 1
         shape = (1, out_channels, int(width), int(height))
 
-        layer = SubtractiveResetIFNodes(
-            shape=shape, reset=0, thresh=1, refrac=0
-        )
+        if node_type:
```
djsaunde: Same here.

bindsnet/conversion/__init__.py
```diff
@@ -469,7 +475,7 @@ def _ann_to_snn_helper(prev, current):
 
 
 def ann_to_snn(ann: Union[nn.Module, str], input_shape: Sequence[int], data: Optional[torch.Tensor] = None,
-               percentile: float = 99.9) -> Network:
+               percentile: float = 99.9, node_type: Optional[nodes.Nodes] = None, **kwargs) -> Network:
```
djsaunde: Similarly here

PR #193, opened 2019-02-17
add missing super.forward in Adaptive nodes

sinkingsugar:
It seems that `AdaptiveLIFNodes` were missing superclass forward call.
Necessary to update traces and such.


djsaunde:
Thanks for the catch!

PR #206, opened 2019-02-28
remove redundent cpu().numpy() on numpy tensor

### LINKED ISSUE #200: minimal_mnist.py results in AttributeError: 'numpy.ndarray' object has no attribute 'cpu'
On fresh setup, I get an error when running minimal_mnist example

```
➜ pip install -e .
...

➜ python --version
Python 3.6.5

➜ python minimal_mnist.py
Loading training images from serialized object file.

Loading training labels from serialized object file.

Traceback (most recent call last):
  File "minimal_mnist.py", line 17, in <module>
    encoding=poisson, time=350, plot_interval=1)
  File "/Users/amolk/work/bindsnet/bindsnet/pipeline/__init__.py", line 113, in __init__
    self.plot_data()
  File "/Users/amolk/work/bindsnet/bindsnet/pipeline/__init__.py", line 249, in plot_data
    self.voltage_record, plot_type=self.plot_type, threshold=self.threshold_value
  File "/Users/amolk/work/bindsnet/bindsnet/analysis/plotting.py", line 413, in plot_voltages
    v[1].cpu().numpy()[n_neurons[v[0]][0]:n_neurons[v[0]][1], time[0]:time[1]].cpu().numpy().T
AttributeError: 'numpy.ndarray' object has no attribute 'cpu'
```

amolk:
Resolves #200

### DIFF
--- a/bindsnet/analysis/plotting.py
+++ b/bindsnet/analysis/plotting.py
@@ -410,7 +410,7 @@ def plot_voltages(voltages: Dict[str, torch.Tensor], ims: Optional[List[AxesImag
                 if plot_type == 'line':
                     ims.append(
                         axes[i].plot(
-                            v[1].cpu().numpy()[n_neurons[v[0]][0]:n_neurons[v[0]][1], time[0]:time[1]].cpu().numpy().T
+                            v[1].cpu().numpy()[n_neurons[v[0]][0]:n_neurons[v[0]][1], time[0]:time[1]].T
                         )
                     )
                     if threshold is not None:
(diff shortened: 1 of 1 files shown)


bindsnet/analysis/plotting.py
```diff
@@ -410,7 +410,7 @@ def plot_voltages(voltages: Dict[str, torch.Tensor], ims: Optional[List[AxesImag
                 if plot_type == 'line':
                     ims.append(
                         axes[i].plot(
-                            v[1].cpu().numpy()[n_neurons[v[0]][0]:n_neurons[v[0]][1], time[0]:time[1]].cpu().numpy().T
+                            v[1].cpu().numpy()[n_neurons[v[0]][0]:n_neurons[v[0]][1], time[0]:time[1]].T
```
djsaunde: This won't work when `spikes` maps from `str`s to `torch.Tensor`s. A better idea would be to convert all input data to `numpy` arrays at the beginning of the function, if they're not already.

amolk: Not sure I understand. `v1[1]` is `torch.tensor` that gets converted to `numpy` array with `v[1].cpu().numpy()`. Then `v[1].cpu().numpy()[...]` is already a `numpy` array.

Converting `voltages` values to `numpy` at the start will probably be more efficient.

djsaunde: Oh I see, that makes sense.

PR #207, opened 2019-02-28
Fix typo n_inpt -> n_input

amolk:
No functionality change.
**Warning**: This change would break existing apps that use `n_inpt` param name


Huizerd:
Why is this considered a typo @djsaunde? Actually, it makes things more inconsistent with the rest of the repository, since the use of `inpt` and `inpts` as variable names is much more common than the use of `input`.

djsaunde:
@Huizerd yes, I agree. Let's change it back.

PR #213, opened 2019-03-18
Correct weight clamping and voltage line plotting

### LINKED ISSUE #212: Weight clamping ineffective
`Connection` gives you the possibility of clamping your weights to a range `wmin`-`wmax`: https://github.com/Hananel-Hazan/bindsnet/blob/f640fb602c6b550b4c9858ee533848fc3c02f135/bindsnet/network/topology.py#L136

However, this clamping doesn't seem to be very effective, as it happens in the `super` call to `LearningRule`:
https://github.com/Hananel-Hazan/bindsnet/blob/f640fb602c6b550b4c9858ee533848fc3c02f135/bindsnet/learning/__init__.py#L55

and weights are used in `Network.get_inputs`, which calls `Connection.compute`:
https://github.com/Hananel-Hazan/bindsnet/blob/f640fb602c6b550b4c9858ee533848fc3c02f135/bindsnet/network/topology.py#L173

This in itself wouldn't be a problem, if it weren't for the fact that weights can be increased/decreased beyond `wmin`/`wmax` between the clamping and the actual use, see for example here:
https://github.com/Hananel-Hazan/bindsnet/blob/f640fb602c6b550b4c9858ee533848fc3c02f135/bindsnet/learning/__init__.py#L535

Is this intended behaviour, am I missing something, or does this indeed seem wrong? I can do something about this, but I was wondering about your preferred way of solving this.

Huizerd:
Resolves #212 and complements #206 (2 lines with identical problems were forgotten there). I added an `assert` to prevent weights from being outside of `wmin`-`wmax` again.

### DIFF
--- a/bindsnet/learning/__init__.py
+++ b/bindsnet/learning/__init__.py
@@ -124,7 +124,6 @@ class PostPre(LearningRule):
         """
         Post-pre learning rule for ``Connection`` subclass of ``AbstractConnection`` class.
         """
-        super().update()
 
         source_s = self.source.s.view(-1).float()
         source_x = self.source.x.view(-1)
@@ -139,12 +138,13 @@ class PostPre(LearningRule):
         if self.nu[1]:
             self.connection.w += self.nu[1] * torch.ger(source_x, target_s)
 
+        super().update()
+
     def _conv2d_connection_update(self, **kwargs) -> None:
         # language=rst
         """
         Post-pre learning rule for ``Conv2dConnection`` subclass of ``AbstractConnection`` class.
         """
-        super().update()
 
         # Get convolutional layer parameters.
         out_channels, _, kernel_height, kernel_width = self.connection.w.size()
@@ -169,6 +169,8 @@ class PostPre(LearningRule):
         if self.nu[1]:
             post = s_target @ x_source.t()
             self.connection.w += self.nu[1] * post.view(self.connection.w.size())
+
+        super().update()
 
 
 class WeightDependentPostPre(LearningRule):
@@ -213,7 +215,6 @@ class WeightDependentPostPre(LearningRule):
         """
         Post-pre learning rule for ``Connection`` subclass of ``AbstractConnection`` class.
         """
-        super().update()
 
         source_s = self.source.s.view(-1).float()
         source_x = self.source.x.view(-1)
@@ -232,12 +233,13 @@ class WeightDependentPostPre(LearningRule):
 
         self.connection.w += update
 
+        super().update()
+
     def _conv2d_connection_update(self, **kwargs) -> None:
         # language=rst
         """
         Post-pre learning rule for ``Conv2dConnection`` subclass of ``AbstractConnection`` class.
         """
-        super().update()
 
         # Get convolutional layer parameters.
         out_channels, in_channels, kernel_height, kernel_width = self.connection.w.size()
@@ -266,6 +268,8 @@ class WeightDependentPostPre(LearningRule):
             update += self.nu[1] * post.view(self.connection.w.size()) * (self.wmax - self.connection.wmin)
 
         self.connection.w += update
+
+        super().update()
 
 
 class Hebbian(LearningRule):
@@ -304,7 +308,6 @@ class Hebbian(LearningRule):
         """
         Hebbian learning rule for ``Connection`` subclass of ``AbstractConnection`` class.
         """
-        super().update()
 
         source_s = self.source.s.view(-1).float()
         source_x = self.source.x.view(-1)
@@ -320,12 +323,13 @@ class Hebbian(LearningRule):
             source_x, target_s
         )
 
+        super().update()
+
     def _conv2d_connection_update(self, **kwargs) -> None:
         # language=rst
         """
         Hebbian learning rule for ``Conv2dConnection`` subclass of ``AbstractConnection`` class.
         """
-        super().update()
 
         out_channels, _, kernel_height, kernel_width = self.connection.w.size()
         padding, stride = self.connection.padding, self.connection.stride
@@ -347,6 +351,8 @@ class Hebbian(LearningRule):
         # Post-synaptic update.
         post = s_target @ x_source.t()
         self.connection.w += self.nu[1] * post.view(self.connection.w.size())
+
+        super().update()
 
 
 class MSTDP(LearningRule):
@@ -390,7 +396,6 @@ class MSTDP(LearningRule):
         :param float a_plus: Learning rate (post-synaptic).
         :param float a_minus: Learning rate (pre-synaptic).
         """
-        super().update()
 
         source_s = self.source.s.view(-1).float()
         source_x = self.source.x.view(-1)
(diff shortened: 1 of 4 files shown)


djsaunde:
Actually, the assert fails when `wmin` or `wmax` is `None`. @Huizerd could you look into this? I think having defaults of `-np.inf` and `np.inf` might fix the problem.

Huizerd:
I also fixed a problem (it looked like a problem) with `SparseConnection`, where it wasn't actually checked whether `wmin` and `wmax` were not `None` before. And I figured out how to test before committing :)

bindsnet/learning/__init__.py
```diff
@@ -194,7 +196,7 @@ def __init__(self, connection: AbstractConnection, nu: Optional[Union[float, Seq
         )
 
         assert self.source.traces, 'Pre-synaptic nodes must record spike traces.'
-        assert connection.wmin is not None and connection.wmax is not None, 'Connection must define wmin and wmax.'
+        assert connection.wmin != -np.inf and connection.wmax != np.inf, 'Connection must define finite wmin and wmax.'
```
djsaunde: We allow infinite weights. This assert should be removed.

bindsnet/network/topology.py
```diff
@@ -148,12 +148,12 @@ def __init__(self, source: Nodes, target: Nodes, nu: Optional[Union[float, Seque
 
         self.w = kwargs.get('w', None)
         if self.w is None:
-            if self.wmin is None or self.wmax is None:
+            if self.wmin == -np.inf or self.wmax == np.inf:
```
djsaunde: What is one is finite and the other isn't? E.g., if `self.wmax` is 0.5, `torch.rand()` may give values larger than it.

bindsnet/network/topology.py
```diff
@@ -148,12 +148,12 @@ def __init__(self, source: Nodes, target: Nodes, nu: Optional[Union[float, Seque
...
                 self.w = torch.rand(source.n, target.n)
-            elif self.wmin is not None and self.wmax is not None:
+            else:
                 self.w = self.wmin + torch.rand(source.n, target.n) * (self.wmax - self.wmin)
         else:
-            if self.wmin is not None and self.wmax is not None:
+            if self.wmin != -np.inf and self.wmax != np.inf:
```
djsaunde: What is one is finite? E.g, if `self.wmin = -np.inf` and `self.wmax=1`, this statement won't execute. Should be an `or` clause.

djsaunde:
This is great work! I made a few comments about using `or` instead of `and` clauses when checking for infinite `wmin`, `wmax`.

bindsnet/learning/__init__.py
Huizerd: Yes we do in general, but for this specific learning rule (which is based on the difference between weight bounds and actual weights), they should be finite right? Otherwise weights would become infinite here:
https://github.com/Hananel-Hazan/bindsnet/blob/f640fb602c6b550b4c9858ee533848fc3c02f135/bindsnet/learning/__init__.py#L227

bindsnet/network/topology.py
```diff
@@ -641,10 +652,15 @@ def __init__(self, source: Nodes, target: Nodes, nu: Optional[Union[float, Seque
...
-            v = self.wmin + (self.wmax - self.wmin) * torch.rand(*source.shape, *target.shape)[i.byte()]
+            if self.wmin == -np.inf or self.wmax == np.inf:
+                v = torch.clamp(torch.rand(*source.shape, *target.shape)[i.byte()], self.wmin, self.wmax)
+            else:
+                v = self.wmin + torch.rand(*source.shape, *target.shape)[i.byte()] * (self.wmax - self.wmin)
             self.w = torch.sparse.FloatTensor(i.nonzero().t(), v)
-        elif self.w is not None:
+        elif self.w is not None and self.sparsity is None:
             assert self.w.is_sparse, 'Weight matrix is not sparse (see torch.sparse module)'
+            if self.wmin != -np.inf or self.wmax != np.inf:
```
Huizerd: I think one should check for weight bounds here as well

bindsnet/network/topology.py
```diff
@@ -562,11 +571,13 @@ def __init__(self, source: Nodes, target: Nodes, nu: Optional[Union[float, Seque
         super().__init__(source, target, nu, weight_decay, **kwargs)
 
         self.w = kwargs.get('w', None)
-
         if self.w is None:
-            self.w = (torch.randn(1)[0] + 1) / 10
+            if self.wmin == -np.inf or self.wmax == np.inf:
```
Huizerd: This also needs clamping/rescaling, to keep consistent with other connections?

bindsnet/network/topology.py
```diff
@@ -273,9 +273,17 @@ def __init__(self, source: Nodes, target: Nodes, kernel_size: Union[int, Tuple[i
 
         assert target.shape[1] == shape[1] and target.shape[2] == shape[2] and target.shape[3] == shape[3], error
 
-        self.w = kwargs.get('w', torch.rand(self.out_channels, self.in_channels, *self.kernel_size))
-        if self.wmin is not None and self.wmax is not None:
-            self.w = torch.clamp(self.w, self.wmin, self.wmax)
+        self.w = kwargs.get('w', None)
+        if self.w is None:
+            if self.wmin == -np.inf or self.wmax == np.inf:
```
Huizerd: Now works the same as for `Connection`, I think the old one only did clamping, even in cases where both `wmin` and `wmax` are defined (so scaling seems more appropriate)

bindsnet/network/topology.py
djsaunde: Yup.

bindsnet/network/topology.py
djsaunde: Yup.

bindsnet/learning/__init__.py
djsaunde: No, I think the weights should be able to get arbitrarily large. They won't actually reach `np.inf` / `-np.inf` since that's impossible.

bindsnet/network/topology.py
djsaunde: Cool, thanks.

Hananel-Hazan:
Great work @Huizerd @djsaunde, and thank you @Huizerd for your PR

djsaunde:
@Huizerd can you open a new PR? This one was mistakenly merged in early.

Huizerd:
Yes I will, I'm still looking at your comment on the assert.

bindsnet/learning/__init__.py
Huizerd: Ok, so say the assert would be removed, and `self.wmin` would be set to `connection.wmin`, which would be `-np.inf`. Then, when learning, `self.update` would be called (which would be `self._connection_update` for a `Connection`) and a proposed `update` to the weights would be computed based on the difference `self.connection.w - self.wmin`. The `update` would become `inf` because the difference is infinite. And, therefore, `self.connection.w` would become `inf`.

According to this, this update rule just won't work without the weight bounds, and it maybe would be better to let people know, rather than let them find out their weights have gone to infinity.

If anything is wrong with the above, please explain because I really don't see it :sweat_smile:

djsaunde: Ah, okay, that makes sense. I think this is all set to merge, then.

PR #216, opened 2019-03-20
Better scaling for reward plot + axis labels

Huizerd:
In case accumulated rewards are constant for some episodes, the horizontal line is obscured since it's right at the figures edge when using `min` and `max` for `set_ylim`. Instead, we could do automatic scaling with `relim` and `autoscale_view`, which works well and leaves margins. See for example https://stackoverflow.com/questions/10984085/automatically-rescale-ylim-and-xlim-in-matplotlib. If interested, I could also add the option to plot a moving average of reward instead of the actual rewards, since those tend to be very shaky.

### DIFF
--- a/bindsnet/pipeline/__init__.py
+++ b/bindsnet/pipeline/__init__.py
@@ -223,19 +223,16 @@ class Pipeline:
     def plot_reward(self) -> None:
         # language=rst
         """
-        Plot the change of accumulated reward for each episodes
+        Plot the accumulated reward for each episode.
         """
         if self.reward_im is None and self.reward_ax is None:
-            fig, self.reward_ax = plt.subplots()
+            self.reward_im, self.reward_ax = plt.subplots()
             self.reward_ax.set_title('Reward')
             self.reward_plot, = self.reward_ax.plot(self.reward_list)
         else:
-            reward_array = np.array(self.reward_list)
-            y_min = reward_array.min()
-            y_max = reward_array.max()
-            self.reward_ax.set_xlim(left=0, right=self.episode)
-            self.reward_ax.set_ylim(bottom=y_min, top=y_max)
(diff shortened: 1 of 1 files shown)


djsaunde:
Looks good. Which would you prefer, the actual reward, or the moving average? Could you possible add an argument to `Pipeline`'s constructor to choose?

Huizerd:
Yes, that's what I was thinking. Would it be ok to use a `pandas` method for this? I saw it's already in the `requirements.txt` anyway, and it's the fastest method for something like this

djsaunde:
Yeah, that's fine. We may want to remove the pandas dependency in the future, but for now it's okay.

bindsnet/pipeline/__init__.py
```diff
@@ -225,14 +228,24 @@ def plot_reward(self) -> None:
...
+        else:
+            reward_list_ = self.reward_list[:]
+
         if self.reward_im is None and self.reward_ax is None:
             self.reward_im, self.reward_ax = plt.subplots()
             self.reward_ax.set_title('Accumulated reward')
             self.reward_ax.set_xlabel('Episode')
             self.reward_ax.set_ylabel('Reward')
-            self.reward_plot, = self.reward_ax.plot(self.reward_list)
+            self.reward_plot, = self.reward_ax.plot(reward_list_)
```
djsaunde: If `self.reward_window` is `None`, is `reward_list_` created?

Huizerd: Yes, that is done in the above if-else statement. If `self.reward_window` is `None`, `reward_list_ = self.reward_list[:]`.

djsaunde: Nice, I missed that for some reason.

PR #219, opened 2019-03-28
MSTDP(ET) implementation

### LINKED ISSUE #217: Implementation of MSTDP and MSTDPET
I'm currently working on reward-modulated STDP, and noticed that there are some discrepancies between the Florian 2007 paper and your MSTDP(ET) implementation. This was already mentioned in #140 and fixed (it seems) in #141, however these changes were reverted in #165. Was this intentional? I was planning on fixing everything with some PRs, but if there are reasons why you wouldn't want the original paper implementation in BindsNET and thus reverted the fixes, I would like to know :smile:

Huizerd:
Implementation of MSTDP/MSTDPET and LIF nodes from the paper by R. Florian (2007). Should I include one of the experiments as example as well? Resolves #217.


djsaunde:
You can include the experiment if you'd like, but it may need maintenance as BindsNET changes. Alternatively, you can add it to the [experiments](https://github.com/djsaunde/bindsnet_experiments) repo.

bindsnet/learning/__init__.py
```diff
@@ -374,7 +386,8 @@ def __init__(self, connection: AbstractConnection, nu: Optional[Union[float, Seq
             connection=connection, nu=nu, weight_decay=weight_decay, **kwargs
         )
 
-        assert self.source.traces and self.target.traces, 'Both pre- and post-synaptic nodes must record spike traces.'
+        assert not self.source.traces and not self.target.traces, \
```
djsaunde: You can just remove this `assert`. There may be other reasons for recording spike traces.

bindsnet/learning/__init__.py
```diff
@@ -385,46 +398,58 @@ def __init__(self, connection: AbstractConnection, nu: Optional[Union[float, Seq
...
+        :param tc_minus: Time constant for post-synaptic firing trace.
         """
 
+        # Reshape pre- and post-synaptic spikes.
         source_s = self.source.s.view(-1).float()
-        source_x = self.source.x.view(-1)
         target_s = self.target.s.view(-1).float()
-        target_x = self.target.x.view(-1)
 
+        # TODO: check this if necessary
```
djsaunde: I think it is, because `w` needs to be 2D since `self.eligibility` is 2D. But you could just reshape `self.eligibility` before using it on line 436, which might be cleaner.

bindsnet/learning/__init__.py
```diff
@@ -493,7 +518,8 @@ def __init__(self, connection: AbstractConnection, nu: Optional[Union[float, Seq
             connection=connection, nu=nu, weight_decay=weight_decay, **kwargs
         )
 
-        assert self.source.traces and self.target.traces, 'Both pre- and post-synaptic nodes must record spike traces.'
+        assert not self.source.traces and not self.target.traces, \
```
djsaunde: Remove.

bindsnet/network/__init__.py
```diff
@@ -303,9 +303,7 @@ def run(self, inpts: Dict[str, torch.Tensor], time: int, **kwargs) -> None:
 
             # Run synapse updates.
             for c in self.connections:
-                self.connections[c].update(
-                    mask=masks.get(c, None), learning=self.learning, **kwargs
-                )
+                self.connections[c].update(dt=self.dt, mask=masks.get(c, None), learning=self.learning, **kwargs)
```
djsaunde: I think I prefer the old formatting:
```
self.connections[c].update(
    dt=self.dt, mask=masks.get(c, None), learning=self.learning, **kwargs
)
```

bindsnet/learning/__init__.py
```diff
@@ -119,10 +119,12 @@ def __init__(self, connection: AbstractConnection, nu: Optional[Union[float, Seq
                 'This learning rule is not supported for this Connection type.'
             )
 
-    def _connection_update(self, **kwargs) -> None:
+    def _connection_update(self, dt: float, **kwargs) -> None:
```
dee0512: The dt parameter can move to the _init_ function of the LearningRule class instead of adding it to all the _connection_update functions to prevent repetition.

djsaunde: Actually, I like this approach better. It's like this for `Nodes` subclasses. That way, you can call `Network.run()` with different `dt` on different simulations.

dee0512: I don't understand. You mean network and connection update will have different dt?

djsaunde: I mean, when you call `Network.run()`, you specify `dt` as an argument. This `dt` is used in `Nodes`, so it should also be used in `Connection`s and `LearningRule`s subclasses. This is somewhat contrived, but you could do something like:

```
network.run(inpts=inpts, dt=1.0, ...)  # Run with 1ms timestep
network.run(inpts=inpts, dt=0.1, ...)  # Run with 0.1ms timestep
```

You can use this to check for numerical instability.

If you passed in `dt` in `__init__` functions, you run the risk of accidentally assigning different `dt`s to different parts of the network, which would be a problem.

dee0512: If we have
`connection.dt = self.dt`
in the `network.add_connection function`, the update rule should have the same dt as the connection.

djsaunde: Yeah, that would work. You could do that for `Nodes` and `LearningRule` objects, too. But then you'd want a function `Network.set_dt()` which propagated changes to child objects.

dee0512: I mean we already have this line in the network.add_connection function:
`connection.dt = self.dt`

And the following line in the network.add_layer function:
`layer.dt = self.dt`

So a simple thing would be to have : 
`connection.update_rule.dt = self.dt` below that. Thus preventing the need to pass a dt every time we want to update the connection and adding an additional parameter to each update_connection function. It would also automatically set the same dt for everything.

Hananel-Hazan:
We currently don't have support for different dt . It will create a
problematic situation where one part of the network excute one iteration as
1ms and other part will execute the same iteration on different dt. It will
create unsync steps between parts of the network

djsaunde:
> We currently don't have support for different dt . It will create a problematic situation where one part of the network excute one iteration as 1ms and other part will execute the same iteration on different dt. It will create unsync steps between parts of the network

We are not proposing multiple `dt`s.

bindsnet/learning/__init__.py
djsaunde: Ah, I forgot about that. Sorry, I've been away from the codebase from a while. So let's have `self.update_rule.dt = connection.dt` in the `__init__` function of the `LearningRule` subclasses?

dee0512: Hmm, the problem is that the connection.dt is set when it is added to a network. So then the update_rule.dt will be none.

bindsnet/learning/__init__.py
Huizerd: I checked, and `self.connection.w` already is 2D. This is confirmed by the fact that the `.view` is missing in the `MSTDPET` rule, but it still functions.

https://github.com/Hananel-Hazan/bindsnet/blob/646dd7de73a1658bb622755e8bb3ebc406a265cd/bindsnet/learning/__init__.py#L542

bindsnet/learning/__init__.py
Huizerd: So, what is the preferred approach here?

djsaunde: Let's implement Devdhar's suggestion: use ‘self.connection.dt‘ in the update methods instead of passing it in to the learning rules' constructor.

Huizerd:
Two more questions I have:
- I tried to convert the `_conv2d_connection_update` variants as well, but didn't really understand the way the kernels etc. are implemented. Do you want me to change these too, and if so, could you help out?
- I was wondering what the units of the time constants used everywhere are. I took them to be milliseconds, cause that's the way the papers do it, but I also saw this comment: https://github.com/Hananel-Hazan/bindsnet/blob/646dd7de73a1658bb622755e8bb3ebc406a265cd/bindsnet/network/__init__.py#L91

djsaunde:
> I tried to convert the _conv2d_connection_update variants as well, but didn't really understand the way the kernels etc. are implemented. Do you want me to change these too, and if so, could you help out?

Yeah, I can take a look at the `_conv2d_connection_update` variants once this is merged in. Unless you want to give it a try.

> I was wondering what the units of the time constants used everywhere are. I took them to be milliseconds, cause that's the way the papers do it...

Really, the `dt` is dimensionless, but we usually refer to them as milliseconds and have designed the rest of the package around it. E.g., decay rates of voltages, spike traces, etc. are all at "biologically plausible" scales in milliseconds. So, we usually communicate about `dt` in terms of milliseconds, but we take the real-time implications with a grain of salt. That is, the simulations probably need lots of tuning before they well-approximate real neurons / synapses.

Huizerd:
Ok, thanks. Would be great if you could look at the convolutions. I'll change the doc string right away.

bindsnet/network/topology.py
```diff
@@ -297,12 +303,14 @@ def compute(self, s: torch.Tensor) -> torch.Tensor:
         """
         return F.conv2d(s.float(), self.w, self.b, stride=self.stride, padding=self.padding, dilation=self.dilation)
 
-    def update(self, **kwargs) -> None:
+    def update(self, dt: float, **kwargs) -> None:
```
djsaunde: Remove `dt` argument here and in other connection update rules.

Huizerd:
@dee0512 You're right, with a different decay the `FlorianNodes` are pretty much similar to the `LIFNodes`. Should I remove them?

Another question: in the `LIFNodes`, the argument `decay` is not used as a time constant (as is written in the arguments description).
https://github.com/Hananel-Hazan/bindsnet/blob/646dd7de73a1658bb622755e8bb3ebc406a265cd/bindsnet/network/nodes.py#L328
Should this be changed?

djsaunde:
@Huizerd it's just a bad doc string. I would simply re-write it to say "Neuron voltage decay rate" or something to that effect.

If you can use the `LIFNodes` and remove the `FlorianNodes` with the MSTDP(ET) demo still working, that would be great.

djsaunde:
@Huizerd I added MSTDP(ET) implementations for `Conv2dConnection`s / did code cleanup / added some tests. Let me know if you have any thoughts.

djsaunde:
Once the `FlorianNodes` are removed, let's get this merged.

Huizerd:
Ok, all set! I will open a new PR for the time constants issue, and I also worked on passing on rewards.

djsaunde:
Merged! Thanks for the great work.

PR #221, opened 2019-04-01
Make a copy of the nn.Module object for conversion

sharath:
The conversion toolkit shouldn't modify the nn.Module object. There might be a better way to accomplish the same thing though.

I also removed the print statements for the time elapsed.

### DIFF
--- a/bindsnet/conversion/__init__.py
+++ b/bindsnet/conversion/__init__.py
@@ -5,6 +5,7 @@ import torch.nn.functional as F
 
 from torch.nn.modules.utils import _pair
 
+from copy import deepcopy
 from time import time as t
 from typing import Union, Sequence, Optional, Tuple, Dict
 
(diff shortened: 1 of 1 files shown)


sharath:
The conversion turns off autograd in data_based_normalization()
[Here's a demo of what I mean](https://gist.github.com/sharath/3be4eed95ead96cdd1e0d385659a8f10)

dee0512:
@sharath ok. A simple fix is ann = torch.tensor(ann). No need to deepcopy. And why remove the logs?

sharath:
It doesn't seem necessary? I think it makes more sense to issue a RuntimeWarning when data is None actually

dee0512:
I agree that the time elapsed isn't necessary but the "starting data normalization" is helpful since it helps debug when there is an error. I have had my programs stuck after that data normalization multiple times. Including a warning is good too. It would help inform that the weights are not being scaled.

PR #223, opened 2019-04-02
Correcting the use of time constants

Huizerd:
Instead of specifying decay rates, all decays now use time constants and exponential functions as suggested in https://github.com/Hananel-Hazan/bindsnet/pull/219#issuecomment-478566867. However, it seems that this causes much steeper decay, as can be demonstrated with the following code:
```python
import torch

v = torch.tensor(-55.0)
rest = torch.tensor(-65.0)
dt = torch.tensor(1.0)

print(f'Input voltage: {v}')

# Old method
decay = torch.tensor(1e-2)
v_old = v - dt * decay * (v - rest)

# New method
tc_decay = torch.tensor(1 / decay.item())
v_new = v - torch.exp(-dt / tc_decay) * (v - rest)

print(f'Old resulting voltage: {v_old:.2f}, change: {(v - v_old):.2f}')
print(f'New resulting voltage: {v_new:.2f}, change: {(v - v_new):.2f}')
```
which prints
```bash
Input voltage: -55.0
Old resulting voltage: -55.10, change: 0.10
New resulting voltage: -64.90, change: 9.90
```

So, something needs to be changed! It looks as if the `torch.exp` is only appropriate in the solution to the differential equation of the LIF neuron, see for example https://neuronaldynamics.epfl.ch/online/Ch1.S3.html (eq. 1.5 for the differential equation, eq. 1.6 for its solution). What do you think?

### DIFF
--- a/bindsnet/models/__init__.py
+++ b/bindsnet/models/__init__.py
@@ -37,9 +37,9 @@ class TwoLayerNetwork(Network):
         self.n_neurons = n_neurons
         self.dt = dt
 
-        self.add_layer(Input(n=self.n_inpt, traces=True, trace_tc=5e-2), name='X')
+        self.add_layer(Input(n=self.n_inpt, traces=True, tc_trace=20.0), name='X')
         self.add_layer(LIFNodes(n=self.n_neurons, traces=True, rest=-65.0, reset=-65.0, thresh=-52.0, refrac=5,
-                                decay=1e-2, trace_tc=5e-2), name='Y')
+                                tc_decay=100.0, tc_trace=20.0), name='Y')
 
         w = 0.3 * torch.rand(self.n_inpt, self.n_neurons)
         self.add_connection(Connection(source=self.layers['X'], target=self.layers['Y'], w=w, update_rule=PostPre,
@@ -56,9 +56,9 @@ class DiehlAndCook2015(Network):
 
     def __init__(self, n_inpt: int, n_neurons: int = 100, exc: float = 22.5, inh: float = 17.5, dt: float = 1.0,
                  nu: Optional[Union[float, Sequence[float]]] = (1e-4, 1e-2), wmin: float = 0.0, wmax: float = 1.0,
-                 norm: float = 78.4, theta_plus: float = 0.05, theta_decay: float = 1e-7,
-                 X_Ae_decay: Optional[float] = None, Ae_Ai_decay: Optional[float] = None,
-                 Ai_Ae_decay: Optional[float] = None) -> None:
+                 norm: float = 78.4, theta_plus: float = 0.05, tc_theta_decay: float = 1e7,
+                 tc_X_Ae_decay: Optional[float] = None, tc_Ae_Ai_decay: Optional[float] = None,
+                 tc_Ai_Ae_decay: Optional[float] = None) -> None:
         # language=rst
         """
         Constructor for class ``DiehlAndCook2015``.
@@ -73,10 +73,10 @@ class DiehlAndCook2015(Network):
         :param wmax: Maximum allowed weight on input to excitatory synapses.
         :param norm: Input to excitatory layer connection weights normalization constant.
         :param theta_plus: On-spike increment of ``DiehlAndCookNodes`` membrane threshold potential.
-        :param theta_decay: Time constant of ``DiehlAndCookNodes`` threshold potential decay.
-        :param X_Ae_decay: Decay of activation of connection from input to excitatory neurons.
-        :param Ae_Ai_decay: Decay of activation of connection from excitatory to inhibitory neurons.
-        :param Ai_Ae_decay: Decay of activation of connection from inhibitory to excitatory neurons.
+        :param tc_theta_decay: Time constant of ``DiehlAndCookNodes`` threshold potential decay.
+        :param tc_X_Ae_decay: Time constant of connection activation decay from input to excitatory neurons.
+        :param tc_Ae_Ai_decay: Time constant of connection activation decay from excitatory to inhibitory neurons.
+        :param tc_Ai_Ae_decay: Time constant of connection activation decay from inhibitory to excitatory neurons.
         """
         super().__init__(dt=dt)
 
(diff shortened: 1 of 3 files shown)


test/network/test_nodes.py
```diff
@@ -34,12 +34,12 @@ def test_init(self):
...
             for n in [1, 100, 10000]:
-                layer = nodes(n, rest=0.0, reset=-10.0, thresh=10.0, refrac=3, decay=7e-4)
+                layer = nodes(n, rest=0.0, reset=-10.0, thresh=10.0, refrac=3, tc_decay=7e4)
 
                 assert layer.rest == 0.0;
                 assert layer.reset == -10.0;
                 assert layer.thresh == 10.0
                 assert layer.refrac == 3;
-                assert layer.decay == 7e-4
+                assert layer.tc_decay == 7e4
```
djsaunde: 1 / 7e-4 = 1428.57...

bindsnet/models/__init__.py
```diff
@@ -86,28 +86,29 @@ def __init__(self, n_inpt: int, n_neurons: int = 100, exc: float = 22.5, inh: fl
...
+        self.add_layer(Input(n=self.n_inpt, traces=True, tc_trace=20.0), name='X')
         self.add_layer(DiehlAndCookNodes(n=self.n_neurons, traces=True, rest=-65.0, reset=-60.0, thresh=-52.0, refrac=5,
-                                         decay=1e-2, trace_tc=5e-2, theta_plus=theta_plus, theta_decay=theta_decay),
+                                         tc_decay=100.0, tc_trace=20.0, theta_plus=theta_plus,
+                                         tc_theta_decay=tc_theta_decay),
                        name='Ae')
 
-        self.add_layer(LIFNodes(n=self.n_neurons, traces=False, rest=-60.0, reset=-45.0, thresh=-40.0, decay=1e-1,
-                                refrac=2, trace_tc=5e-2),
+        self.add_layer(LIFNodes(n=self.n_neurons, traces=False, rest=-60.0, reset=-45.0, thresh=-40.0, tc_decay=100.0,
```
djsaunde: 1 / 1e-1 = 10

Huizerd:
I also found some time constants in the `DiehlAndCook2015` model which aren't actually used:
https://github.com/Hananel-Hazan/bindsnet/blob/727ed38d97716b21bfbaeb3f8d17da5bc809f5b8/bindsnet/models/__init__.py#L60
They are passed to `Connection`s as the keyword argument `decay`, but that keyword is only used for `MaxPool2dConnection`:
https://github.com/Hananel-Hazan/bindsnet/blob/727ed38d97716b21bfbaeb3f8d17da5bc809f5b8/bindsnet/network/topology.py#L381

test/network/test_nodes.py
Huizerd: I was too quick there.. Do you prefer `1/7e-4`, or something else?

djsaunde: Let's approximate it as 1.5e3 for now.

djsaunde: Actually, I thought the decay was `1e-2`? https://github.com/Hananel-Hazan/bindsnet/blob/727ed38d97716b21bfbaeb3f8d17da5bc809f5b8/bindsnet/network/nodes.py#L315

djsaunde:
> I also found some time constants in the DiehlAndCook2015 model which aren't actually used:

Yup, those can be safely deleted. They were part of an experiment that we never followed up on.

test/network/test_nodes.py
Huizerd: True, but I thought you had given it a different value here for testing purposes

djsaunde:
> So, something needs to be changed! It looks as if the torch.exp is only appropriate in the solution to the differential equation of the LIF neuron, see for example https://neuronaldynamics.epfl.ch/online/Ch1.S3.html (eq. 1.5 for the differential equation, eq. 1.6 for its solution). What do you think?

According to equation 1.6, you want the following:

```
import torch 
 
v = torch.tensor(-55.0) 
rest = torch.tensor(-65.0) 
dt = torch.tensor(1.0) 
 
print(f'Input voltage: {v}') 
 
# Old method 
decay = torch.tensor(1e-2) 
v_old = v - dt * decay * (v - rest) 
  
# New method 
tc_decay = torch.tensor(1 / decay.item()) 
v_new = torch.exp(-dt / tc_decay) * (v - rest) + rest

print(f'Old resulting voltage: {v_old:.2f}, change: {(v - v_old):.2f}') 
print(f'New resulting voltage: {v_new:.2f}, change: {(v - v_new):.2f}')                                                                     

Input voltage: -55.0
Old resulting voltage: -55.10, change: 0.10
New resulting voltage: -55.10, change: 0.10
```
We have been hard-cording forward (Euler) integration up until this point, hence the incremental `v - dt * decay * (v - rest)` code, whereas (correct me if I'm wrong) `torch.exp(-dt / tc_decay) * (v - rest) + rest` is the exact solution.

djsaunde:
@Huizerd could you also fix the time constants in the scripts in the `examples/` folder?

Huizerd:
> @Huizerd could you also fix the time constants in the scripts in the `examples/` folder?

Yes, of course!

> We have been hard-cording forward (Euler) integration up until this point, hence the incremental `v - dt * decay * (v - rest)` code, whereas (correct me if I'm wrong) `torch.exp(-dt / tc_decay) * (v - rest) + rest` is the exact solution.

Yes, that's what I thought. So you want to go with the solution, or leave it as a differential equation + Euler? I can check for speed differences between the two approaches, but I don't know of any other differences between the two.

djsaunde:
> Yes, that's what I thought. So you want to go with the solution, or leave it as a differential equation + Euler? I can check for speed differences between the two approaches, but I don't know of any other differences between the two.

Hm, let's go with the `torch.exp()` solution, if it can be done for all `Nodes` types (can it?). Unless you prefer the other implementation and can convince us it's better. I think the exact solution (?) is easier to read / understand / extend, however.

I expect `torch.exp` to be slightly more expensive than the existing implementation, but the computational bottleneck of our code should remain in `Connection`s computation, so I'm not really worried about it.

Huizerd:
> Hm, let's go with the `torch.exp()` solution, if it can be done for all `Nodes` types (can it?). Unless you prefer the other implementation and can convince us it's better. I think the exact solution (?) is easier to read / understand / extend, however.
> 
> I expect `torch.exp` to be slightly more expensive than the existing implementation, but the computational bottleneck of our code should remain in `Connection`s computation, so I'm not really worried about it.

I'll come back to this.





> @Huizerd could you also fix the time constants in the scripts in the `examples/` folder?

I didn't find any usage of `tc_trace` or `decay` there, all networks use the default values.

djsaunde:
> I didn't find any usage of tc_trace or decay there, all networks use the default values.

Great, thanks for checking anyways.

Huizerd:
Calculating the solution each time was about 1.5x slower than the forward Euler, 12 vs 8 microseconds (so not really an issue). There were differences in the voltages over longer periods of time however, and even though these were small (0.04mV after 0.1s), I think it would be better to go with the exact solution.

djsaunde:
> Calculating the solution each time was about 1.5x slower than the forward Euler, 12 vs 8 microseconds (so not really an issue). There were differences in the voltages over longer periods of time however, and even though these were small (0.04mV after 0.1s), I think it would be better to go with the exact solution.

I agree. Awesome work!

PR #229, opened 2019-04-09
Revert exact integration and neuron decay reparametization.

Hananel-Hazan:
Reverting changes made by @Huizerd.
1. The neuron decay was not intuitive, small decay constant should influence by small changes and not vice versa.
2. revert the neuron update and remove the exp from the activation functions, its make the simulation slower and don't add any value to the activation function  
3. default values break backward compatibility.
4. rename trace_tc to trace. No need to the prefix tc_

### DIFF
--- a/bindsnet/models/__init__.py
+++ b/bindsnet/models/__init__.py
@@ -37,9 +37,9 @@ class TwoLayerNetwork(Network):
         self.n_neurons = n_neurons
         self.dt = dt
 
-        self.add_layer(Input(n=self.n_inpt, traces=True, trace=20.0), name='X')
+        self.add_layer(Input(n=self.n_inpt, traces=True, trace=5e-2), name='X')
         self.add_layer(LIFNodes(n=self.n_neurons, traces=True, rest=-65.0, reset=-65.0, thresh=-52.0, refrac=5,
-                                decay=100.0, trace=20.0), name='Y')
+                                decay=1e-2, trace=5e-2), name='Y')
 
         w = 0.3 * torch.rand(self.n_inpt, self.n_neurons)
         self.add_connection(Connection(source=self.layers['X'], target=self.layers['Y'], w=w, update_rule=PostPre,
@@ -81,14 +81,14 @@ class DiehlAndCook2015(Network):
         self.inh = inh
         self.dt = dt
 
-        self.add_layer(Input(n=self.n_inpt, traces=True, trace=20.0), name='X')
+        self.add_layer(Input(n=self.n_inpt, traces=True, trace=5e-2), name='X')
         self.add_layer(DiehlAndCookNodes(n=self.n_neurons, traces=True, rest=-65.0, reset=-60.0, thresh=-52.0, refrac=5,
-                                         decay=100.0, trace=20.0, theta_plus=theta_plus,
+                                         decay=1e-2, trace=5e-2, theta_plus=theta_plus,
                                          theta_decay=theta_decay),
                        name='Ae')
 
         self.add_layer(LIFNodes(n=self.n_neurons, traces=False, rest=-60.0, reset=-45.0, thresh=-40.0, decay=10.0,
-                                refrac=2, trace=20.0),
+                                refrac=2, trace=5e-2),
                        name='Ai')
 
         w = 0.3 * torch.rand(self.n_inpt, self.n_neurons)
@@ -139,12 +139,12 @@ class DiehlAndCook2015v2(Network):
         self.inh = inh
         self.dt = dt
 
-        input_layer = Input(n=self.n_inpt, traces=True, trace=20.0)
+        input_layer = Input(n=self.n_inpt, traces=True, trace=5e-2)
         self.add_layer(input_layer, name='X')
 
         output_layer = DiehlAndCookNodes(
             n=self.n_neurons, traces=True, rest=-65.0, reset=-60.0, thresh=-52.0, refrac=5,
-            decay=100.0, trace=20.0, theta_plus=theta_plus, theta_decay=theta_decay
+            decay=1e-2, trace=5e-2, theta_plus=theta_plus, theta_decay=theta_decay
         )
         self.add_layer(output_layer, name='Y')
 
@@ -196,12 +196,12 @@ class IncreasingInhibitionNetwork(Network):
         self.max_inhib = max_inhib
         self.dt = dt
 
-        input_layer = Input(n=self.n_input, traces=True, trace=20.0)
+        input_layer = Input(n=self.n_input, traces=True, trace=5e-2)
         self.add_layer(input_layer, name='X')
 
         output_layer = DiehlAndCookNodes(
             n=self.n_neurons, traces=True, rest=-65.0, reset=-60.0, thresh=-52.0, refrac=5,
-            decay=100.0, trace=20.0, theta_plus=theta_plus, theta_decay=theta_decay
+            decay=1e-2, trace=5e-2, theta_plus=theta_plus, theta_decay=theta_decay
         )
         self.add_layer(output_layer, name='Y')
 
@@ -285,13 +285,13 @@ class LocallyConnectedNetwork(Network):
                          int((input_shape[1] - kernel_size[1]) / stride[1]) + 1)
 
         if real:
-            input_layer = RealInput(n=self.n_inpt, traces=True, trace=20.0)
+            input_layer = RealInput(n=self.n_inpt, traces=True, trace=5e-2)
         else:
-            input_layer = Input(n=self.n_inpt, traces=True, trace=20.0)
+            input_layer = Input(n=self.n_inpt, traces=True, trace=5e-2)
 
         output_layer = DiehlAndCookNodes(
             n=self.n_filters * conv_size[0] * conv_size[1], traces=True, rest=-65.0, reset=-60.0,
-            thresh=-52.0, refrac=5, decay=100.0, trace=20.0, theta_plus=theta_plus, theta_decay=theta_decay
+            thresh=-52.0, refrac=5, decay=1e-2, trace=5e-2, theta_plus=theta_plus, theta_decay=theta_decay
         )
         input_output_conn = LocallyConnectedConnection(
             input_layer, output_layer, kernel_size=kernel_size, stride=stride, n_filters=n_filters,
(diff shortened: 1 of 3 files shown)


djsaunde:
I strongly disagree with this PR. I've made arguments below.

> The neuron decay was not intuitive, small decay constant should influence by small changes and not vice versa.

Actually, time constants are typically used this way. See, e.g., BRIAN2, also [here](https://en.wikipedia.org/wiki/Time_constant). They're typically defined in units of time (not time^{-1}, as we have been using). I've been meaning to make this change for a while.

> revert the neuron update and remove the exp from the activation functions, its make the simulation slower and don't add any value to the activation function

As @Huizerd showed, it makes the integration of voltage slightly slower, but the major of simulation time isn't spent on this, so it barely makes a difference in performance. Also, it's exact numerical simulation, whereas we were using an approximate that was inaccurate with certain choices of time constant.

> default values break backward compatibility

Where? There aren't any scripts in the examples/ directory that use custom time constants. The new time constants are simply the inverse of the old time constants, which preserves backwards compatibility.

> rename trace_tc to trace. No need to the prefix tc_

I think prefixing `tc_` to all time constant variables is a good naming convention (helps to understand what they _do_).

Hananel-Hazan:
Thanks @djsaunde. 
We are not Brian, we are not imitating biology, we only inspire from biology. Currently we are supporting only one time constant. 
Starting to change peace by peace its not a good practice. From here we start to get more de-fragmented. If you want to make that change, open a new branch, when it will be ready we merge it.  
Currently, this change brake all the experiments that we running. 

> As @Huizerd showed, it makes the integration of voltage slightly slower, but the major of simulation time isn't spent on this, so it barely makes a difference in performance. Also, it's exact numerical simulation, whereas we were using an approximate that was inaccurate with certain choices of time constant.
It make the code too slow, and the approximation was good enough. When we find a case that its not holding up we will change it. For now, simple is better. 

> default values break backward compatibility
In our running simulation, @dee0512 , @sharath and I break our head to find why our code stop running.

> I think prefixing tc_ to all time constant variables is a good naming convention (helps to understand what they do).
It is, I agree, but when it will be needed.  When in the future we will support multiple time constants, that will be relevant. For now, there is no need for that. its only make the parameters name clunky.

djsaunde:
> We are not Brian, we are not imitating biology, we only inspire from biology. Currently we are supporting only one time constant.
Starting to change peace by peace its not a good practice. From here we start to get more de-fragmented. If you want to make that change, open a new branch, when it will be ready we merge it.
Currently, this change brake all the experiments that we running.

I am not claiming we are BRIAN, or that we're imitating biology. I'm saying that this way of using time constants is the way it is commonly used; i.e., it's good practice.

Which experiments did it break? Did you use custom time constants somewhere? If you did, just replace them with their inverse (1 / time constant).

Moreover, the feature is ready. It was developed in a different branch, and a PR was made, reviewed, and merged. These sorts of improvements can be a little painful, but I believe it will be worth it in the long run.

> It make the code too slow, and the approximation was good enough. When we find a case that its not holding up we will change it. For now, simple is better.

I didn't notice any slowdown; can you quantify the slowdown you're seeing? It seems both updates operate in the ~10ns regime.

Here's a case where the approximation breaks down; namely, when using a fast time course:

```
import torch 
 
v = torch.tensor(-55.0) 
rest = torch.tensor(-65.0) 
dt = torch.tensor(1.0) 
 
print(f'Input voltage: {v}') 

# Old method 
decay = torch.tensor(0.5) 
v_old = v - dt * decay * (v - rest) 
  
# New method 
tc_decay = torch.tensor(1 / decay.item()) 
v_new = v - (1 - torch.exp(-dt / tc_decay)) * (v - rest) 
 
print(f'Old resulting voltage: {v_old:.2f}, change: {(v - v_old):.2f}') 
print(f'New resulting voltage: {v_new:.2f}, change: {(v - v_new):.2f}')                                                                      
Input voltage: -55.0
Old resulting voltage: -60.00, change: 5.00
New resulting voltage: -58.93, change: 3.93
```

> In our running simulation, @dee0512 , @sharath and I break our head to find why our code stop running.

Like I said, use the inverse of the old time constant (1 / old value), like in the above example.

> It is, I agree, but when it will be needed. When in the future we will support multiple time constants, that will be relevant. For now, there is no need for that. its only make the parameters name clunky.

I think you're conflating time constants and the simulation timestep. Different time constants are used in many places: voltage decay, spike trace decay, etc. Hence, the prefix `tc_` to let the user know they are choosing a time constant that governs some time-varying quantity.

Huizerd:
@Hananel-Hazan 

> 2. revert the neuron update and remove the exp from the activation functions, its make the simulation slower and don't add any value to the activation function

While I do agree that the `torch.exp` makes the simulation a bit slower (one of my simulations went from 29min to 35min), numerical accuracy could also be important. As @djsaunde points out, the approximation breaks down for larger time constants (since this just isn't the way to solve a differential equation).

If you're not willing to sacrifice simulation speed for numerical accuracy, then let's at least keep the use of time constants in there: so `dt / time_const` instead of `dt * decay_factor`. This makes the entire repository much more compatible with all papers on neurocomputing, which all make use of time constants (see e.g. [Vasilaki et al., 2009](https://intranet.physio.unibe.ch/Publikationen/Dokumente/Vasilaki2009PloSComputBio_1.pdf), [Florian, 2007](https://florian.io/papers/2007_Florian_Modulated_STDP.pdf), [Fremaux et al., 2010](http://www.jneurosci.org/content/jneuro/30/40/13326.full.pdf), etc.).

> 3. default values break backward compatibility.

It might have been frustrating to see your program run and give very different results, but this is a one time issue which pays off in the long run (for the reasons mentioned above).

> 4. rename trace_tc to trace. No need to the prefix tc_

> It is, I agree, but when it will be needed. When in the future we will support multiple time constants, that will be relevant. For now, there is no need for that. its only make the parameters name clunky.

In the `nodes.py` file alone there are multiple time constants in use for different neurons: `tc_trace` for decaying the trace, `tc_decay` for voltage leaking of leaky neurons, `tc_theta_decay` for threshold decay, `tc_i_decay` for current decay. In the `__init__.py` of the `learning` module there are even more: `tc_plus` and `tc_minus` for decaying pre- and post-synaptic firing traces, for instance. So there are actually many time constants, and it would be in your interest to indicate they are time constants.

dee0512:
If the decay is just inverse, why not fix just make it inverse in the equation instead of requiring all the existing scripts to change their decay to 1/decay? This is how it would backward compatible with the previous values. Backward compatibility does not require changing existing code. We spent 2 days figuring out what is wrong in our scripts and it turned out that our decay values were too high because of the new implementation. 

The Florian and RL experiments use much smaller networks than the ones we usually work with on bindsnet. Therefore, without benchmark on the time it adds for bigger networks (some of our experiments already take 7+ hours), we should not change it. 


In the following example:
```
import torch 
 
v = torch.tensor(-55.0) 
rest = torch.tensor(-65.0) 
dt = torch.tensor(1.0) 
 
print(f'Input voltage: {v}') 

# Old method 
decay = torch.tensor(0.5) 
v_old = v - dt * decay * (v - rest) 
  
# New method 
tc_decay = torch.tensor(1 / decay.item()) 
v_new = v - (1 - torch.exp(-dt / tc_decay)) * (v - rest) 
 
print(f'Old resulting voltage: {v_old:.2f}, change: {(v - v_old):.2f}') 
print(f'New resulting voltage: {v_new:.2f}, change: {(v - v_new):.2f}')

```

The difference of value is not due to approximation error but because the following is not true:

` x = 1 - exp(-1/x)`

If I wanted the same amount of decay for the old code, I would choose the decay value as: `0.8646647167633873 (1 - exp(-2))` and not 0.5. And I would get the same result but faster.

djsaunde:
> If the decay is just inverse, why not fix just make it inverse in the equation instead of requiring all the existing scripts to change their decay to 1/decay? This is how it would backward compatible with the previous values. Backward compatibility does not require changing existing code. We spent 2 days figuring out what is wrong in our scripts and it turned out that our decay values were too high because of the new implementation.

Backwards compatibility is broken sometimes. Yes, it's a pain, but this change should be beneficial in the long run, since __specifying time constants in units of time is common practice__. This will help us communicate better with other researchers. It just looks bad to use parameters with incorrect units.

> The Florian and RL experiments use much smaller networks than the ones we usually work with on bindsnet. Therefore, without benchmark on the time it adds for bigger networks (some of our experiments already take 7+ hours), we should not change it.

Yeah, that's fair. I ran a simple benchmark: `examples/mnist/eth_mnist.py` before and after the change, and the exact solution takes ~0.05s more per example, which is pretty significant.

In light of this, I think we should just revert the `torch.exp` change (back to Euler method), but keep the inverted time constants.

> The difference of value is not due to approximation error but because the following is not true:
x = 1 - exp(-1/x)

I'm not sure I understand this point. The error is in the approximation to the LIF neuron model.

> If I wanted the same amount of decay for the old code, I would choose the decay value as: 0.8646647167633873 (1 - exp(-2)) and not 0.5. And I would get the same result but faster.

Yeah, you could do that, but that's shifting the burden of simulation onto the user of BindsNET. And, you only know to do that because you're familiar with the simulation internals.

All in all, I think the Euler method approximation should be good enough, especially with large time constants, so we could revert that. But we should definitely keep the inverted time constants. Good libraries occasionally break backwards compatibility for the sake of better design.

dee0512:
Ok I agree.

If you want to keep the exp function:
`self.v = self.rest + torch.exp(-self.dt / self.tc_decay) * (self.v - self.rest)`

you can do something like :
`self.calculated_decay = torch.exp(-self.dt / self.tc_decay) `

so that it is done once when the node is created and not every step. I believe this can be done everywhere.

Huizerd:
> Ok I agree.
> 
> If you want to keep the exp function:
> `self.v = self.rest + torch.exp(-self.dt / self.tc_decay) * (self.v - self.rest)`
> 
> you can do something like in the init function :
> `self.calculated_decay = torch.exp(-self.dt / self.tc_decay) `
> 
> so that it is done once when the node is created and not every step. I believe this can be done everywhere.

Great idea! I could make a pull request for this if you guys want.

djsaunde:
Yeah, that sounds good. Let's wait for the okay from @Hananel-Hazan, then @Huizerd can make the PR.

Hananel-Hazan:
Yes, I agree.
Great work and great discussion everyone.

djsaunde:
Great! @Huizerd could you put together a PR?

PR #240, opened 2019-05-03
Migrate im2col_indices and col2im_indices to pytorch functions

k-chaney:
The previous implementation passed tensors to numpy and a transfer from GPU to CPU would have to take place if the tensor originally live on the GPU.

bindsnet/utils.py
- im2col_indices - This was changed from a custom implementation to the implementation found in pytorch in torch.nn.functional as the unfold operation. im2col is just a special case of unfold. This was tested by comparing exact values in place using examples/mnist/conv_mnist.py

- col2im_indices - This was changed from a custom implementation to the implementation found in pytorch in torch.nn.functional as the fold operation. col2im is just a special case of fold.

### DIFF
--- a/bindsnet/utils.py
+++ b/bindsnet/utils.py
@@ -3,52 +3,17 @@ import torch
 import numpy as np
 
 from torch import Tensor
+import torch.nn.functional as F
 from numpy import ndarray
 from typing import Tuple, Union
 from torch.nn.modules.utils import _pair
-
-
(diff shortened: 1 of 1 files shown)


djsaunde:
This looks great. I'll look into this and try to get it merged ASAP.

djsaunde:
@Hananel-Hazan looks good to me; I recommend merging.

Hananel-Hazan:
Thanks @k-chaney for the PR

Issue #242, opened 2019-05-05
Shape of tensors during convolutions


k-chaney:
For Conv2dConnection, since it utilizes pytorch's conv2d operator it is inherently constrained by the same input expectations, which is BxCxHxW.

Inside of examples/mnist/conv_mnist.py the shape of the tensors go as follows
Line 120 - images has shape - torch.Size([60000, 28, 28])
Line 125 - images is fed to poisson_loader which effectively turns it into - torch.Size([60000, 50, 28, 28])
Line 147 - a single spike stream is pulled out with shape - torch.Size([50, 28, 28])
               - this is then formatted through two unsqueezes which becomes shape -  torch.Size([50, 1, 1, 28, 28])
Which plays nicely with the sampling in bindsnet/network/__init__.py (line 286) which turns it into torch.Size([1,1,28,28]) and in turn is the shape that conv2d expects.

However this exact formula breaks down with the CIFAR10 dataset which constructs a tensor of shape - torch.Size([50000, 32, 32, 3]) - and thus transpositions must be done.

This complexity isn't mirrored in a non-convolutional network (such as the one defined in examples/cifar10/supervised_cifar10.py).

My proposal would be to standardize the output from the supported datasets to work with both setups.

My suggestion would be to follow in pytorch's footsteps and leverage their data loading pipeline and after the DataLoader produce TxBxCxHxW. Ultimately for larger datasets (such as MVSEC, DDD17, and others) this paradigm will be significantly more flexible.

One concern would be breaking some of the backwards compatibility in other portions of the code (this can be helped a bit through keeping the primary access functions around and just deprecate them for removal later).

I'm more than happy to work on this and can probably have something for review by Tuesday or Wednesday depending on my exact schedule.

djsaunde:
Yeah, I think this is a reasonable thing to do. It might make sense to directly leverage `torchvision.datasets` (instead of `bindsnet.datasets`; or `bindsnet.datasets` would lightly wrap it?) so we no longer have to maintain things like MNIST, CIFAR-10, etc. and benefit from datasets they support which we currently do not.

Hananel-Hazan:
Merged #243

PR #243, opened 2019-05-07
[WIP] Standardizing datasets

k-chaney:
So this is my initial take on #242. There is a light wrapper around all of the datasets provided by torchvision which should be relatively easy to maintain.

`spike_encoders.py` houses wrappers around your encoding that are setup to work like the transforms provided by torchvision. Should this live in `bindsnet/datasets` or should it live in `bindsnet/encoding` since these are relatively generic objects?

The changes to `bindsnet/network/__init__.py` is motivated by keeping the default pytorch DataLoader which creates (in our case) BxTxCxHxW and having the time dimension be arbitrary seems like a good solution.

Not everything is implemented yet (hence WIP), but I wanted to make sure this was heading in a direction you guys are comfortable with before spending the time to wrap everything up. The TODO list includes putting spoken_mnist back in, example for a non-convolutional network, update other examples to use this new loading paradigm, work in a dataset that originated as spiking (simply to provide an example from that direction).


bindsnet/datasets/torchvision_wrapper.py
```diff
@@ -0,0 +1,58 @@
...
+        def __init__(self,
+                     image_encoder: Encoder,
+                     label_encoder: Optional[Encoder],
+                     *args, **kwargs):
+            """
+            Constructor for the BindsNET torchvision dataset wrapper.
+            For details on the dataset you're interested in visit
+
+            https://pytorch.org/docs/stable/torchvision/datasets.html
+
```
djsaunde: Delete this line

bindsnet/network/__init__.py
```diff
@@ -231,6 +231,8 @@ def run(self, inpts: Dict[str, torch.Tensor], time: int, **kwargs) -> None:
         :param Union[float, torch.Tensor] reward: Scalar value used in reward-modulated learning.
         :param Dict[Tuple[str], torch.Tensor] masks: Mapping of connection names to boolean masks determining which
                                                      weights to clamp to zero.
+        :param Union[int, Dict[str, int]] input_time_dim: Dimension for slicing in time for inputs. Can vary
```
djsaunde: Is this necessary? Can it just always be 0?

bindsnet/datasets/torchvision_wrapper.py
```diff
@@ -0,0 +1,58 @@
+from typing import Optional
+
+import torchvision
+
+from .spike_encoders import Encoder
+
+def vd_wrapper_creator(ds_type):
```
djsaunde: What does `vd` stand for?

bindsnet/datasets/__init__.py
```diff
@@ -1,945 +1,9 @@
...
-                d['data'].append(temp[b'data'].reshape(-1, 3, 32, 32).transpose(0, 2, 3, 1))
-                d['labels'].append(temp[b'fine_labels'])
-
-        data, labels = np.concatenate(d['data']), np.concatenate(d['labels'])
-        return data.astype(np.float32), labels.astype(np.float32)
+MNIST = vd_wrapper_creator('MNIST')
+CIFAR10 = vd_wrapper_creator('CIFAR10')
+CIFAR100 = vd_wrapper_creator('CIFAR100')
+FashionMNIST = vd_wrapper_creator('FashionMNIST')
+ImageFolder = vd_wrapper_creator('ImageFolder')
```
djsaunde: Very cool. Can you add the rest of the PyTorch datasets?

bindsnet/datasets/__init__.py
```diff
@@ -1,945 +1,9 @@
...
 
+from .spike_encoders import *
```
djsaunde: Why import everything here? Might want to do `__all__` = ['spike_encoder']`, which allows Python to find the submodule. See e.g. [here](https://github.com/Hananel-Hazan/bindsnet/blob/master/bindsnet/__init__.py).

bindsnet/datasets/spike_encoders.py
```diff
@@ -0,0 +1,91 @@
+import torch
+from typing import Tuple, List, Dict
+
+from .. import encoding
+
+__all__ = ['Encoder', 'SingleEncoder', 'RepeatEncoder',
+           'BernoulliEncoder', 'PoissonEncoder', 'RankOrderEncoder']
+
```
djsaunde: Add line here

bindsnet/network/__init__.py
```diff
@@ -283,7 +293,13 @@ def run(self, inpts: Dict[str, torch.Tensor], time: int, **kwargs) -> None:
             for l in self.layers:
                 # Update each layer of nodes.
                 if isinstance(self.layers[l], AbstractInput):
-                    self.layers[l].forward(x=inpts[l][t])
+                    # grab the time slice for each input
```
djsaunde: I'm not sure what these lines do, could you explain?

djsaunde:
I really like the `Encoder` abstraction, and that it uses the already existing `bindsnet.encoding` functions. I'm interested to see how this works out!

bindsnet/network/__init__.py
k-chaney: This was the simplest way I could think of to deal with the consequences of the pytorch DataLoader and default_collate_fn which will make everything BxTxCxWxH. Essentially we would have to maintain our own collate_fn and notation for which tensors should have the batch as the first dimension and which shouldn't.

bindsnet/datasets/torchvision_wrapper.py
k-chaney: VisionDataset -- this is something that they're working on adding. I was reading their source code when putting this together and not their docs. I'll switch it to something more meaningful for the current implementation in torchvision.

bindsnet/datasets/__init__.py
k-chaney: Yes, I'll add the rest, just wanted a small subset to test with initially. Should I make a note for which ones I've tested and which ones I haven't? Some I won't be able to (e.g. cityscapes and other large ones)

bindsnet/datasets/__init__.py
k-chaney: I can change it to that.

bindsnet/datasets/__init__.py
Hananel-Hazan: That is a good idea

djsaunde: Yes, could you please create a `README.md` in the `datasets/` folder which has these notes?

bindsnet/network/__init__.py
k-chaney: Essentially, to deal with the fact that the time dimension could be anywhere I keep around a dictionary of slices for each input separately (different shape, time dimension, etc.) and I fill in the time into that list at the appropriate dimension. `slice(None)` is the same as `:`. I hope that helps clarify a bit.

djsaunde: Yup, thanks.

k-chaney:
So I've been working on this more and this has pretty far reaching consequences inside of Pipeline, Environment, and all of the examples. I will keep moving on this, my only question revolves around backwards compatibility. There are some locations that have pretty hard dependencies to maintain long term such as:

https://github.com/Hananel-Hazan/bindsnet/blob/4021360cc0f715947bf6dc3f7effab003141cb4c/bindsnet/environment/__init__.py#L169

https://github.com/Hananel-Hazan/bindsnet/blob/4021360cc0f715947bf6dc3f7effab003141cb4c/bindsnet/environment/__init__.py#L260

https://github.com/Hananel-Hazan/bindsnet/blob/4021360cc0f715947bf6dc3f7effab003141cb4c/bindsnet/pipeline/__init__.py#L214


@djsaunde What're your thoughts regarding the overall impact and if I should put a lot of effort into maintaining backwards compatibility? My plan as of right now is to create more generic base classes and then provide examples of sub classing them. This would wind up breaking current users' code to some degree, but it should be relatively small changes on their end.

djsaunde:
> What're your thoughts regarding the overall impact and if I should put a lot of effort into maintaining backwards compatibility?

The hard dependencies that you mentioned we definitely don't want in the library. I'd say, feel free to break backwards compatibility here. If you can come up with an abstraction that will allow us to render datasets / environments using the `Pipeline` object, that would be great; having to write custom code for each is not a good approach.

k-chaney:
@djsaunde I want to check on the function of the history component inside of the pipeline as I've been racking my brain for a little bit on this. It looks like it overwrites self.obs before passing it into the encoder to use a delta from a previous time as an observation. This seems like it would be a dataset or environment dependent component.

djsaunde:
I'm on vacation at the moment; @dee0512 might be good to ask about this!

dee0512:
@k-chaney I just checked the pipeline. I agree, it should be part of dataset transformation rather than part of the pipeline if that is what you are asking.

k-chaney:
@dee0512 Yes, that's what I was asking about. Everything is approaching a good state at this point. I just need to reintegrate the plotting utilities into the pipeline, add comments, and update examples (how many of the examples should I port?). Would you mind doing a quick read through to make sure everything makes sense?

djsaunde:
@k-chaney I'm back from vacation; hopefully I'll have some time today to do a read-through and add some comments.

bindsnet/environment/__init__.py
```diff
@@ -126,12 +126,22 @@ def step(self, a: int) -> Tuple[torch.Tensor, float, bool, Dict[Any, Any]]:
             self.update_history()
             self.update_index()
 
-        self.obs = self.encoder(self.obs)
+        raw_obs = self.obs
+
+        # The new standard is BxTxCxHxW. The gym environment doesn't
+        # follow exactly the same protocol.
+        self.obs = self.encoder(self.obs).unsqueeze(1)
```
djsaunde: Should this be `unsqueeze(0)`?

bindsnet/environment/__init__.py
```diff
@@ -126,12 +126,22 @@ def step(self, a: int) -> Tuple[torch.Tensor, float, bool, Dict[Any, Any]]:
...
+        if self.obs.dim() == 4:
+            self.obs = self.obs.unsqueeze(2)
 
         self.episode_step_count += 1
 
         # Return converted observations and other information.
-        return self.obs, self.reward, self.done, info
+        return {"obs": self.obs,
```
djsaunde: Why put this in a dictionary? This breaks the Gym environment standard

bindsnet/pipeline/environment_pipeline.py
```diff
@@ -115,8 +117,10 @@ def env_step(self):
         # Accumulate reward
         self.accumulated_reward += self.reward
 
-        return {'obs': self.obs, 'reward': reward,
-                'done': self.done, 'info': info}
+        batch['reward'] = self.reward
```
djsaunde: Should return obs (state), reward, done, info

bindsnet/environment/__init__.py
k-chaney: I've rewritten this part to be a bit more clear

bindsnet/environment/__init__.py
k-chaney: Okay, I read up on the Gym environment standard and am now adding extra useful debug info (such as the original gym observation) to the info dictionary.

bindsnet/environment/__init__.py
```diff
@@ -121,27 +121,35 @@ def step(self, a: int) -> Tuple[torch.Tensor, float, bool, Dict[Any, Any]]:
...
+
         # Store frame of history and encode the inputs.
         if len(self.history) > 0:
             self.update_history()
             self.update_index()
-
-        raw_obs = self.obs
+            # add the delta observation into the info for debugging and
+            # display
+            info['delta_obs'] = self.obs
```
djsaunde: Aren't you just storing `self.obs` twice?

bindsnet/environment/__init__.py
```diff
@@ -121,27 +121,35 @@ def step(self, a: int) -> Tuple[torch.Tensor, float, bool, Dict[Any, Any]]:
...
 
         # The new standard is BxTxCxHxW. The gym environment doesn't
         # follow exactly the same protocol.
-        self.obs = self.encoder(self.obs).unsqueeze(1)
-        if self.obs.dim() == 4:
-            self.obs = self.obs.unsqueeze(2)
+
+        # we want CxHxW. If it is in HxW add in  the channel dimension
```
djsaunde: What if the data is 1D?

bindsnet/pipeline/__init__.py
```diff
@@ -1,8 +1,10 @@
 from .environment_pipeline import EnvironmentPipeline
 from .base_pipeline import BasePipeline
 from .dataloader_pipeline import DataLoaderPipeline, TorchVisionDatasetPipeline
-from .pipeline_analysis import PipelineAnalyzer
+
+from . import pipeline_analysis
```
djsaunde: I don't think you need to import these; as long as they're in `__all__`

bindsnet/environment/__init__.py
k-chaney: I pulled the history code that was inside of Pipeline previously. It overwrites `self.obs` internally. So the second will be the delta observation instead of the raw.

bindsnet/environment/__init__.py
k-chaney: Updated the code to include comments on exact assumptions as well as adding a flag in the constructor to not automatically add the channel dimension for 2D observations

bindsnet/pipeline/__init__.py
k-chaney: I think that the latest commit will reflect what you're looking for here.

examples/mnist/conv_mnist.py
```diff
@@ -23,15 +26,15 @@
...
-parser.add_argument("--n_train", type=int, default=60000)
+parser.add_argument("--n_epochs", type=int, default=1)
 parser.add_argument("--n_test", type=int, default=10000)
 parser.add_argument("--kernel_size", type=int, default=16)
 parser.add_argument("--stride", type=int, default=4)
 parser.add_argument("--n_filters", type=int, default=25)
 parser.add_argument("--padding", type=int, default=0)
 parser.add_argument("--time", type=int, default=50)
 parser.add_argument("--dt", type=int, default=1.0)
-parser.add_argument("--intensity", type=float, default=0.5)
```
Hananel-Hazan: Is there is a reason why to change the default values?

examples/mnist/conv_mnist.py
```diff
@@ -23,15 +26,15 @@
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--seed", type=int, default=0)
-parser.add_argument("--n_train", type=int, default=60000)
+parser.add_argument("--n_epochs", type=int, default=1)
```
Hananel-Hazan: Is this should be 60000 like in the `--n_train`?

k-chaney: `--n_train` controls how many iterations. `--n_epochs` will control how many times through the dataset. Which for mnist (if using the standard train and test split) it will be 50k training steps for every epoch.

examples/mnist/conv_mnist.py
k-chaney: Bindsnet's previous implementation of MNIST provided values 0-255 while Pytorch's provides values 0-1. Hence the change from intensity 0.5 to 128, so that the values fed into the PoissonEncoder is the same.

Hananel-Hazan: Thanks

examples/mnist/conv_mnist.py
djsaunde: 60K, but I like this.

djsaunde:
This looks like it's reaching a stable state. @k-chaney: Any more changes to be made before I make one last pass?

k-chaney:
@djsaunde I don't have any more large changes for this pull request

Hananel-Hazan:
Thank you @k-chaney for the Great work!

PR #244, opened 2019-05-08
Fix for bias=False

sharath:
The converted network fails to run when the bias for a layer is initialized as False. This fixes it.

Example:
```
import torch
import torch.nn as nn
from bindsnet.conversion import ann_to_snn

class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = nn.Linear(1, 2, bias=False)
        self.fc2 = nn.Linear(2, 1, bias=False)

    def forward(self, x):
        x = torch.relu(self.fc1(x))
        x = self.fc2(x)
        return x

ann = Net()
snn = ann_to_snn(ann, input_shape=(1,1))
snn.run({'Input': torch.rand(200,1)}, time=200)
```

Output:
`Traceback (most recent call last):
  File "test.py", line 18, in <module>
    snn.run({'Input': torch.rand(200,1)}, time=200)
  File "/Users/sharath/bindsnet/bindsnet/network/__init__.py", line 279, in run
    inpts.update(self.get_inputs())
  File "/Users/sharath/bindsnet/bindsnet/network/__init__.py", line 211, in get_inputs
    inpts[c[1]] += self.connections[c].compute(source.s)
  File "/Users/sharath/bindsnet/bindsnet/network/topology.py", line 180, in compute
    post = s.float().view(-1) @ self.w + self.b
TypeError: add(): argument 'other' (position 1) must be Tensor, not NoneType
`


djsaunde:
Good catch! Could you just change [this line](https://github.com/Hananel-Hazan/bindsnet/blob/4021360cc0f715947bf6dc3f7effab003141cb4c/bindsnet/network/topology.py#L297), though?

`self.b = kwargs.get('b', torch.zeros(self.out_channels))` should be `self.b = kwargs.get('b', torch.zeros(self.n))` (correct me if I'm wrong).

sharath:
Shouldn't it be target.n?

djsaunde:
Why doesn't `target.n` work?

djsaunde:
> Why doesn't target.n work?

Bump

sharath:
It's failing all of the TestLearningRules tests because of the Conv2DConnection change.

I'm getting this error: `RuntimeError:` Given weight of size 32 1 3 3, expected bias to be 1-dimensional with 32 elements, but got bias of size [2048] instead`

From the pytorch documentation for torch.nn.functional.conv2d:

>     bias: optional bias tensor of shape :math:`(\text{out\_channels})`. Default: ``None``

This makes sense since the weights and biases are shared for each of the feature maps.

PR #259, opened 2019-05-28
Disables adaptive threshold when network.learning = False; removing unused code.

djsaunde:
Now, there is no need to manually set theta_decay and theta_plus to zero when testing a network with a layer with adaptive thresholds.

### DIFF
--- a/bindsnet/network/nodes.py
+++ b/bindsnet/network/nodes.py
@@ -666,7 +666,8 @@ class AdaptiveLIFNodes(Nodes):
         """
         # Decay voltages and adaptive thresholds.
         self.v = self.decay * (self.v - self.rest) + self.rest
-        self.theta *= self.theta_decay
+        if self.network.learning:
+            self.theta *= self.theta_decay
 
         # Integrate inputs.
         self.v += (self.refrac_count == 0).float() * x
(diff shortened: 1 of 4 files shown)


bindsnet/network/nodes.py
```diff
@@ -682,7 +683,8 @@ def forward(self, x: torch.Tensor) -> None:
         # Refractoriness, voltage reset, and adaptive thresholds.
         self.refrac_count.masked_fill_(self.s, self.refrac)
         self.v.masked_fill_(self.s, self.reset)
-        self.theta += self.theta_plus * self.s.float()
+        if self.network.learning:
```
Hananel-Hazan: Using the condition in the forward method will slow the performance.
Its rather to check the condition one time and not twice on the same function.

We need to find other way to differentiate between training or testing, maybe two forward function? 
Or when setting `network.learning = False` some values will be zero or one depend on the usage.

bindsnet/network/nodes.py
```diff
@@ -803,7 +806,8 @@ def forward(self, x: torch.Tensor) -> None:
...
+            self.theta += self.theta_plus * self.s.float()
```
Hananel-Hazan: Same here, Prefer to use one condition then two

bindsnet/network/nodes.py
djsaunde:
```%timeit if True: pass                                                   
5.54 ns ± 0.0233 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)

Hananel-Hazan: How about a function to switch between learning and testing? 
Then some parameters can be change between modes

djsaunde: Maybe, but that will require some thought + more code changes. I think this is good for the time being, but we can schedule the train / test switching function for future work.

PR #266, opened 2019-05-31
Device managment

k-chaney:
Leverage `torch.nn.Module` to handle easy transfers between devices and not have to worry about setting default tensor locations.

- `examples/mnist/conv_mnist.py` contains a GPU option that leverages this
- `BasePipeline` will now handle transfer to GPU if it is available by default (both the network and the batch)
- `bindsnet/pipeline/base_pipeline.py` contains a `recursive_to` function that handles transferring an arbitrary batch structure to the target device


bindsnet/network/network.py
```diff
@@ -100,7 +100,8 @@ def __init__(
         self.layers = {}
         self.connections = {}
         self.monitors = {}
-        self.learning = learning
+        self.train(learning)
```
djsaunde: What's your reason for changing this?

bindsnet/network/nodes.py
```diff
@@ -34,6 +35,7 @@ def __init__(
         :param tc_trace: Time constant of spike trace decay.
         :param trace_scale: Scaling factor for spike trace.
         :param sum_input: Whether to sum all inputs.
+        :param learning: Whether to be in learning or testing
```
djsaunde: Period at the end of this comment.

bindsnet/learning/__init__.py
```diff
@@ -71,8 +71,8 @@ def update(self) -> None:
         if (
             self.connection.wmin != -np.inf or self.connection.wmax != np.inf
         ) and not isinstance(self, NoOp):
-            self.connection.w = torch.clamp(
-                self.connection.w, self.connection.wmin, self.connection.wmax
+            self.connection.w.clamp_(
```
djsaunde: Nice!

bindsnet/network/__init__.py
```diff
@@ -1,362 +1,4 @@
-import tempfile
-from typing import Dict, Optional
-
-import torch
-
-from .monitors import AbstractMonitor
-from .nodes import AbstractInput, Nodes
-from .topology import AbstractConnection
+from .network import Network, load
 from ..learning.reward import AbstractReward
```
djsaunde: Is this needed here?

bindsnet/network/topology.py
```diff
@@ -363,15 +369,13 @@ def normalize(self) -> None:
         Normalize weights along the first axis according to total weight per target neuron.
         """
         if self.norm is not None:
-            shape = self.w.size()
-            self.w = self.w.view(
+            # get a view and modify in place
```
djsaunde: Awesome!

bindsnet/network/network.py
k-chaney: For backwards compatibility `self.learning` is kept throughout the code and `self.train()` (the method from `torch.nn.Module`) has been overloaded to include the writing of `self.learning` as these effectively do the same operation. We can use `network.train()` or `network.eval()` as pytorch uses them, however directly writing to `network.training` no longer works. Pytorch internally keeps track of `self.training`. We could switch completely without too much work.

djsaunde: Yeah, that's a good point.

Hananel-Hazan: What Pytorch internally keeps track of when using `self.training`?
How that it help us? beside been more compatible with PyTorch API?

bindsnet/network/topology.py
```diff
@@ -157,21 +157,24 @@ def __init__(
         """
         super().__init__(source, target, nu, weight_decay, **kwargs)
 
-        self.w = kwargs.get("w", None)
-        if self.w is None:
+        w = kwargs.get("w", None)
```
Hananel-Hazan: Is there a redone to this use of internal self.w and temporal w?

bindsnet/network/network.py
k-chaney: https://github.com/pytorch/pytorch/blob/181792176d46b5e14b4d4c63292afb6670f5e27e/torch/nn/modules/module.py#L987

it is used for batchnorm and dropout mainly. When switching between training and testing people are used to calling `train()` and `eval()` on their models. It is just a function that sets the flag in every `torch.nn.Module` that is a child of the primary one that it was called from.

bindsnet/network/topology.py
k-chaney: I wanted to set it as a `Parameter` and you can't pass `None` into that, only `torch.Tensor`s. Thus for the initial logic of creating `self.w` this is done without assigning to `self.w` and instead creating a temporary `w` that doesn't have the same constraints.

PR #269, opened 2019-06-03
Fixing errors related to environment pipelines, plotting configuration

Huizerd:
A lot of things related to environments and pipelines weren't working anymore, so I fixed some errors and added a plotting configuration allowing you to specify options and plotting per plot type (needs to be implemented in other pipelines, just a proof of concept). I still have some questions regarding what to do with `time` (used in `encodings` and `Network.run()`, for example) and `dt`:
1. `time` is an `int`, while `dt` is a `float`, which seems strange
2. `EnvironmentPipeline.step_()` assumed the `time` for `Network.run()` was to be equal to the size of the time dimension of the encoded observation (see https://github.com/Hananel-Hazan/bindsnet/blob/1a89bd1299e931d6d7f2d9aa6efb9cb727ceb1a5/bindsnet/pipeline/environment_pipeline.py#L158). However, this doesn't work in case you have set `Network.dt=10.0` and the `time` argument to your encoding is `10` as well (implying you just want to do everything in steps of 10 ms instead of the standard 1 ms), since then `timesteps=int(time/self.dt)` in `Network.run()` will be `0`.


djsaunde:
Yeah, I suppose `time` could be a float. We'd just have to go through the codebase and make sure it doesn't break anything that assumes it's an `int`. We'll have to use `int(time / dt)` everywhere that this occurs.

bindsnet/pipeline/base_pipeline.py
```diff
@@ -101,35 +99,35 @@ def __init__(self, network: Network, **kwargs):
...
 
         if torch.cuda.is_available() and self.allow_gpu:
-            self.device = torch.device('cuda')
+            self.device = torch.device("cuda")
         else:
-            self.device = torch.device('cpu')
+            self.device = torch.device("cpu")
 
-        self.network.to(self.device)
+        # self.network.to(self.device)
```
djsaunde: Why comment this out?

bindsnet/pipeline/environment_pipeline.py
```diff
@@ -79,42 +74,38 @@ def __init__(
...
+        Runs for the specified number of episodes. Each episode can be an arbitrary length.
         """
-
-        for self.episode in range(self.num_episodes):
+        while self.episode < self.num_episodes:
             self.reset_()
 
-            for step in itertools.count():
+            for _ in itertools.count():
                 batch = self.env_step()
```
djsaunde: This should probably just be unpacked as `state, reward, done, info`, like usual.

examples/space_invaders/minimal_space_invaders.py
```diff
@@ -38,19 +38,23 @@
...
     "SpaceInvaders-v0",
-    BernoulliEncoder(time=1, dt=network.dt),
+    BernoulliEncoder(time=int(network.dt), dt=network.dt),
     history_length=2,
     delta=4,
 )
 environment.reset()
 
+# Plotting configuration.
+plot_config = {"data_step": 1, "data_length": 10, "reward_eps": 1, "reward_window": 10, "volts_type": "line"}
```
djsaunde: I like this refactoring!

bindsnet/pipeline/environment_pipeline.py
```diff
@@ -79,42 +74,38 @@ def __init__(
...
+    def train(self, **kwargs) -> None:
+        # language=rst
         """
-        Runs for the specified number of episodes from the Environment.
-        Each episode can be an arbitrary length.
+        Runs for the specified number of episodes. Each episode can be an arbitrary length.
         """
-
-        for self.episode in range(self.num_episodes):
+        while self.episode < self.num_episodes:
```
Hananel-Hazan: This is unnecessary, whats wrong with for?

examples/cartpole/minimal_cartpole.py
```diff
@@ -0,0 +1,149 @@
+from functools import partial
```
Hananel-Hazan: Nice!

.travis.yml
```diff
@@ -1,7 +1,8 @@
 language: python
+dist: xenial
```
Hananel-Hazan: Is that necessary?

k-chaney: This was also added in the branch that I submitted recently. It is a good practice since we don't know when travis-ci will change their default. Their default (14.04 LTS) is now out of support and thus we shouldn't expect package updates to continue. My thoughts on stuff like this is to always be on the latest LTS (and then if possible test one back as well) that is available (18.04 however I don't think travis-ci has support for it yet).

bindsnet/pipeline/environment_pipeline.py
Huizerd: The `for` loop doesn't allow you to use the `self.episode` anywhere else, so there would be no point in making it an attribute of `self`. Besides, I made use of `self.episode` in `EnvironmentPipeline.plots()`, in order to take care of plotting reward only so many episodes.

bindsnet/pipeline/environment_pipeline.py
Huizerd: Wouldn't that be superfluous? It's easier and more clear (in my opinion) to move all these things between the many functions. Besides, I don't think it's done like that anywhere else.

djsaunde: IMO, it makes the code more readable.

Huizerd: Like this?

PR #271, opened 2019-06-06
Data fixes

coopersigrist:
A refactoring for eth_mnist example. Also set the default value for encoders in torchvision wrapper to None so that we mustn't always specify.


djsaunde:
@coopersigrist can you rebase to the master branch?

bindsnet/datasets/torchvision_wrapper.py
```diff
@@ -31,8 +31,8 @@ class torchvision_dataset_wrapper(ds_type):
 
         def __init__(
             self,
-            image_encoder: Optional[Encoder],
-            label_encoder: Optional[Encoder],
+            image_encoder: Optional[Encoder] = None,
+            label_encoder: Optional[Encoder] = None,
```
k-chaney: I don't think this can be `None` as the order of the arguments matter in this case. The point is to split the requirements. First we take care of Bindsnet requirements `image_encoder` and `label_encoder` then we take care of specific dataset requirements `*args` and `**kwargs`. I think forcing these (even if somebody puts `None`) keeps the consistency. @djsaunde thoughts?

PR #284, opened 2019-06-19
 added davis dataset with working downloader

coopersigrist:
Also a quick script to load the dataset to test download

At the moment the dataset returns a sequence of images and masks which may not be ideal for implementation but we will see.


bindsnet/datasets/davis.py
```diff
@@ -0,0 +1,254 @@
...
+                raise FileNotFoundError(f"Images for sequence {seq} not found.")
+            self.sequences[seq]["images"] = images
+            masks = np.sort(glob(os.path.join(self.mask_path, seq, "*.png"))).tolist()
+            masks.extend([-1] * (len(images) - len(masks)))
+            self.sequences[seq]["masks"] = masks
+
+        # Creates an enumeration for the sequences for __getitem__
+        self.enum_sequences = self.sequences.items()
+
+    # Returns the number of sequences in the dataset
```
djsaunde: Put these comments in doc strings

examples/datasets/Davis.py
```diff
@@ -0,0 +1,10 @@
+import argparse
```
djsaunde: rename to `davis.py`

bindsnet/datasets/davis.py
```diff
@@ -0,0 +1,254 @@
...
+        speed = int(progress_size / (1024 * duration))
+        percent = min(int(count * block_size * 100 / total_size), 100)
+        sys.stdout.write(
+            "\r...%d%%, %d MB, %d KB/s, %d seconds passed"
+            % (percent, progress_size / (1024 * 1024), speed, duration)
+        )
+        sys.stdout.flush()
+
+
+if __name__ == "__main__":
```
Hananel-Hazan: The file need to be an interface, main should not need to be here

djsaunde:
I requested two minor changes, but it looks good overall. Have you tested the dataset locally?

bindsnet/datasets/davis.py
djsaunde: No, this is okay. Code under `__main__` is only run if this file is run as a script.

Hananel-Hazan: Main here is odd. Its not appear at any of the dataset files, why this one is any different?
The part can be on examples.

djsaunde: Fair enough. I agree with Hananel.

coopersigrist:
I'm testing it out now, there is a an issue in the indexing of the sequences, but the download works locally, I tried a few different tests and they all seemed to work

djsaunde:
@coopersigrist Okay, just let us know when you've worked out the issue before we merge.

bindsnet/datasets/davis.py
Hananel-Hazan: Agree, we need to find a way to test the dataset, although this one could be difficult because of the its size.

coopersigrist: Yeah I was originally using the main to test, but I started moving over to the actual example script and I'll remove the main for the next commit

coopersigrist:
got it, I'll make the changes you suggested and figure out whats going on.

bindsnet/datasets/davis.py
djsaunde: Yes, I think that is the point of having this `__main__` code. We can't add dataset tests to Travis CI, they will cause the build to fail (they take too long!).

bindsnet/datasets/davis.py
coopersigrist: do you want me to change all comments to doc strings? Is that our standard?

djsaunde: No, but comments above methods are discouraged. See [this SO post](https://stackoverflow.com/questions/2357230/what-is-the-proper-way-to-comment-functions-in-python). And, and public methods (ones that don't begin with `_`) should have a doc-string.

djsaunde: See also [this PEP](https://www.python.org/dev/peps/pep-0257/).

coopersigrist: ah gotcha, makes sense, I'm on it

bindsnet/datasets/davis.py
```diff
@@ -0,0 +1,254 @@
...
+            yield seq
+
+    # Downloads the dataset
+    def _download(self):
+
+        os.makedirs(self.root)
+
+        # Grabs the correct zip url based on parameters
+        zip_url = (
+            "https://data.vision.ee.ethz.ch/csergi/share/davis/DAVIS-2017-"
```
sharath: Use an f-string to be consistent?

bindsnet/datasets/davis.py
```diff
@@ -0,0 +1,254 @@
...
+            + self.tag
+            + ".zip"
+        )
+
+        print("\nDownloading Davis data set from " + zip_url + "\n")
```
sharath: ditto

coopersigrist:
Alright I made the changes and everything seems to work now

bindsnet/datasets/davis.py
```diff
@@ -111,14 +111,28 @@ def __init__(
...
+        calculates the number of sequences the dataset holds
+
+        :Returns the number of sequences in the dataset
+        """
+
         return len(self.sequences)
 
-    # Verifies that the correct dataset is downloaded; downloads if it isn't and download=True
     def _check_directories(self):
+
```
djsaunde: Remove blank lines before doc strings.

coopersigrist: just changed it

Hananel-Hazan:
Seems fine for me

PR #286, opened 2019-06-21
Fix data loader for reservoir example 

Hananel-Hazan:
Add minor issues like printing, progress bar and arguments


Hananel-Hazan:
Yes, its works

PR #288, opened 2019-06-25
Mini-batching implementation (WIP)

### LINKED ISSUE #245: Minibatch processing
Minibatching data in artificial neural network training / inference allows for vast speedups in computation time and reductions in variance of parameter updates. Minibatching in spiking neural networks is inherently more complicated because spiking neurons maintain state (voltages, spikes traces, etc.).

We may be able to implement mini-batching by duplicating neurons at the start of each simulation, processing minibatches of data independently in parallel, and aggregating the results of each independent simulation to make, e.g., weight updates. The process would work something like this:

1. When calling `network.run`, users will pass in data of shape `[batch_size, time, k_1, ..., k_n]` (matching the WIP PR #243).
2. Neurons' state variables will be duplicated along the batch dimension `batch_size` times; when `batch_size = 1`, this reduces to our current implementation.
3. Simulation proceeds as normal, but updates to neuron state variables occur independently across the batch dimension. Updates to weights occur in parallel.

djsaunde:
This PR is meant to stimulate discussion, and is not necessarily a final version. Closes #245.

Simulation is parallelized by creating N copies of neuron state variables (voltages, spike occurrences, refractory period counters, ...) and aggregating changes to adaptive parameters (weights, adaptive thresholds, ...) over the course of simulation. Network now takes an optional argument `batch_size`, to pre-allocate tensors for component state variables.

Inputs to the network now have shape `[time, n_1, ..., n_k]` or `[batch_size, time, n_1, ..., n_k]`. This subsumes the old implementation. If the rank of an input tensor is 2, mini-batch size is assumed to be 1; if the rank of an input tensor is 1, the minibatch size and simulation time are assumed to be 1. If the rank of the input tensor is >= 3, the first dimension is assumed to be the mini-batch size, the second, simulation time, and the rest must match the shape of the corresponding input layer.

Also supports "dynamic batch size", where, if the input data has minibatch size unequal to what was originally passed, the network re-initializes the component state variables that are sensitive to minibatch size (voltages, spike occurrences, refractory period counters, ...)

Breaking changes:
- Monitor objects recordings now have shape `[time, batch_size, n_1, ..., n_k]` to match most other parts of the code. This is a change from the old shape, `[n_1, ..., n_k, time]`.

Desiderata:
- torch.utils.data.DataLoader loads items from our dataset wrappers with shape `[batch_size, time, n_1, ..., n_k]`. I'd prefer this would match the convention `[time, batch_size, n_1, ..., n_k]` in most other parts of the code (CC @k-chaney, any thoughts on this?)
- Aggregation of changes to adaptive parameters (weights, adaptive thresholds, ...) are summed across the minibatch dimension. As we've discussed (@coopersigrist), this makes sense in most cases, but it would be nice to have other aggregation options (e.g., mean, max) which can be passed as an argument to the constructor of a `LearningRule`.

Let me know if I've missed anything, and I'll update this comment accordingly.


k-chaney:
The shape of the tensors is something that will be a recurring theme for us. We should hash it out over a dedicated issue and then make the update package wide. The reason it is ```[batch, time, n_0,...n_m]``` is to simplify how we interact with the ```DataLoader``` whose default collate function will put the batch as the first dimension. We should have a standard and stick with it (even if it means making a couple functions that parity some pytorch ones).

As far as the shape on the ```Nodes```, is there any reason to not just infer the shape at the next call after a ```reset_()```? The only reason this comes up is because we're already doing extra book-keeping for the ```batch_size```.

Also, something for another issue, we may want to think about implementing monitors through keeping lists of views and then only cat when it is requested. It seems wasteful to slice and cat the data at every record even if it's not used.

djsaunde:
> The shape of the tensors is something that will be a recurring theme for us. We should hash it out over a dedicated issue and then make the update package wide. The reason it is [batch, time, n_0,...n_m] is to simplify how we interact with the DataLoader whose default collate function will put the batch as the first dimension. We should have a standard and stick with it (even if it means making a couple functions that parity some pytorch ones).

I prefer `[time, batch, n_1, ..., n_k]`, since it mirrors how PyTorch implements RNNs. Maybe we can rewrite the `DataLoader`'s `collate_fn` to do this.

> As far as the shape on the Nodes, is there any reason to not just infer the shape at the next call after a reset_()? The only reason this comes up is because we're already doing extra book-keeping for the batch_size.

Yeah, this is essentially what is happening. If the shape of the input is changed, the shape of the state variables changes with it. Maybe I'm misunderstanding you?

> Also, something for another issue, we may want to think about implementing monitors through keeping lists of views and then only cat when it is requested. It seems wasteful to slice and cat the data at every record even if it's not used.

Yes, totally agreed. I was thinking about this while implementing this PR. Could you create an issue for this?

Hananel-Hazan:
Just observation, why dimension places `[time, n_1, ..., n_k]` or `[batch_size, time, n_1, ..., n_k]` differ from thre monitor order of dimension `[time, batch_size, n_1, ..., n_k]`. (batch_size and  time switch places)
Is that intentionally? is there any reason why it will be different? if not, I think it will be good to choose one and stick with it (preferably one that PyTorch already using).

djsaunde:
@Hananel-Hazan Just read the previous comments:

> torch.utils.data.DataLoader loads items from our dataset wrappers with shape [batch_size, time, n_1, ..., n_k]. I'd prefer this would match the convention [time, batch_size, n_1, ..., n_k] in most other parts of the code

Like I said, I want them to match, but the `DataLoader`'s default `collate_fn` collates minibatches of data along the 0-th dimension. To fix this:

> Maybe we can rewrite the DataLoader's collate_fn to do this.

k-chaney:
@djsaunde Do we want to handle the issues surrounding shape in this PR? If so, I can make a collate function that conforms to the ```[time, batch, n_0, ..., n_k]``` standard.

djsaunde:
@k-chaney That would be great if you could. Barring that, I think all the shapes follow the `[time, batch_size, *shape]` convention.

k-chaney:
> > As far as the shape on the Nodes, is there any reason to not just infer the shape at the next call after a reset_()? The only reason this comes up is because we're already doing extra book-keeping for the batch_size.
> 
> Yeah, this is essentially what is happening. If the shape of the input is changed, the shape of the state variables changes with it. Maybe I'm misunderstanding you?
> 

I guess my point is that we are only changing shape along the batch dimension with ```set_batch_size```.  The nodes themselves don't have any inherent reason to not be able to infer the shape (inside of ``forward```) across all dimensions right after a ```reset_()```. In the case of a fully convolutional network, we might want to test on a different size image compared to what we train on.

Essentially this would remove the need to specify an exact shape ahead of time in the nodes (the connections would drive this).

djsaunde:
So, IIUC, you're saying that the shape of the layer itself should be able to change in `forward`?

k-chaney:
Yes, that's the summary

bindsnet/conversion/__init__.py
```diff
@@ -343,27 +308,6 @@ def compute(self, s: torch.Tensor):
...
         return F.pad(s, self.padding).float()
 
-    def update(self, **kwargs) -> None:
-        # language=rst
-        """
-        Dummy definition of abstract method ``update``.
-        """
-        pass
-
-    def normalize(self) -> None:
```
dee0512: I think this will give an error if we do not overwrite the functions.

bindsnet/conversion/__init__.py
```diff
@@ -166,13 +166,6 @@ def reset_(self) -> None:
         self.v = self.reset * torch.ones(self.shape)  # Neuron voltages.
         self.refrac_count = torch.zeros(self.shape)  # Refractory period counters.
 
-    def _compute_decays(self) -> None:
-        # language=rst
-        """
```
dee0512: why remove this?

Hananel-Hazan:
This is obviously need to be tested and require to fix the dependencies of this change, but its approved for merge

djsaunde:
@Hananel-Hazan Agreed. It passes all the tests, and I've written mini-batch versions of `eth_mnist.py` and `conv_mnist.py`, and made sure the original versions worked with the change as well. What remains is to make sure all other examples work with it.

@k-chaney Do you want to implement the collate function we discussed? If so, I'll wait for that before merging.

sharath:
Looks good

k-chaney:
I just pushed up the collate function. I'll include a light wrapper around the DataLoader class which will use this as the default collate. I'll also fix whatever broke from that commit.

Is there a non public forum that I can use to let you guys know some more of my schedule (maybe a slack)? If so just shoot me an email.

djsaunde:
> I just pushed up the collate function. I'll include a light wrapper around the DataLoader class which will use this as the default collate. I'll also fix whatever broke from that commit.

Awesome, thanks.

> Is there a non public forum that I can use to let you guys know some more of my schedule (maybe a slack)? If so just shoot me an email.

We usually communicate over email. Slack might be a good option, though. @Hananel-Hazan Any thoughts? In the meantime, feel free to email us.

Hananel-Hazan:
I understand the hype around Slack, but found it redundant in what it offer. I prefer email or a private Google Group

bindsnet/network/network.py
```diff
@@ -292,13 +292,21 @@ def run(self, inpts: Dict[str, torch.Tensor], time: int, **kwargs) -> None:
         # Dynamic setting of batch size.
         if inpts != {}:
             for key in inpts:
+                # goal shape is [time, batch, n_0, ...]
+
                 if len(inpts[key].size()) == 1:
+                    # current shape is [n_0, ...]
+                    # unsqueeze twice to make [1, 1, n_0, ...]
                     inpts[key] = inpts[key].unsqueeze(0).unsqueeze(0)
```
djsaunde: Does this unsqueeze need to happen? Or can the ops just be broadcast?

djsaunde:
Merged! Thanks everyone.

PR #296, opened 2019-07-17
Fixes Poisson encoder and adds support for arbitrary minibatch reductions in learning rules.

djsaunde:


djsaunde:
@coopersigrist This will give us a chance to test different minibatched parameter update reduction methods for different tasks. You can also implement custom reduction methods, which must take `input` and `dim` arguments (assumed to be a float tensor and integer, respectively), and must output a float tensor. For example, since `torch.max` outputs a 2-tuple of (maximum values, indices), it won't work as a `reduction` argument, but we can modify it to work in a custom function:

```
def max_without_indices(input, dim=0):
    return torch.max(input, dim=dim)[0]

connection = Connection(..., reduction=max_without_indices, ...)
```

bindsnet/network/topology.py
```diff
@@ -623,91 +626,8 @@ def normalize(self) -> None:
...
-        source: Nodes,
-        target: Nodes,
-        nu: Optional[Union[float, Sequence[float]]] = None,
-        weight_decay: float = 0.0,
-        **kwargs
-    ) -> None:
-        # language=rst
-        """
-        Instantiates a :code:`MeanFieldConnection` object.
-
```
coopersigrist: Do we not need MeanFieldConnection anymore?

examples/mnist/batch_eth_mnist.py
```diff
@@ -9,6 +9,7 @@
 
 from time import time as t
 
+from bindsnet import ROOT_DIR
```
coopersigrist: I like this change! Didn't know we had that constant! 

Should we give some option as to the download path or no?

bindsnet/learning/learning.py
```diff
@@ -51,6 +53,12 @@ def __init__(
 
         self.nu = nu
 
+        # Parameter update reduction across minibatch dimension.
```
coopersigrist: all of the reduction changes seem good to me

bindsnet/network/topology.py
Hananel-Hazan: We should retain MeanFieldConnection? The subject going to do a comeback

djsaunde: I don't think we're ever going to use it, but if we do, we can always put it back in (it will always exist in the git history). I don't know if the implementation is correct.

djsaunde: If we need it, we can add it back in. I don't think the implementation is correct, though, so it makes sense to remove it for now. It will still live in the git history, so we can get it back any time.

Hananel-Hazan: I'm using it currently now for Ray/Robert CAN program.
If its not correct we should fix it.

djsaunde: Okay, I'll add it back in!

examples/mnist/batch_eth_mnist.py
djsaunde: I'd rather not add extraneous arguments. We have a lot already, and I like putting it under the project root directory as a default because it houses the data with the source code.

PR #297, opened 2019-07-18
Izhikevich nouron issue 

Hananel-Hazan:
Fix input dimensions issue
Remove unrelevant line

### DIFF
--- a/bindsnet/network/nodes.py
+++ b/bindsnet/network/nodes.py
@@ -1218,7 +1218,7 @@ class IzhikevichNodes(Nodes):
 
         # Add inter-columnar input.
         if self.s.any():
-            x += self.S[:, self.s.view(-1)].sum(dim=1)
+            x += torch.cat([self.S[:, self.s[i]].sum(dim=1)[None] for i in range(self.s.shape[0])], dim=0)
 
         # Apply v and u updates.
         self.v += self.dt * 0.5 * (0.04 * self.v ** 2 + 5 * self.v + 140 - self.u + x)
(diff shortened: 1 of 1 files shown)


bindsnet/network/nodes.py
```diff
@@ -1250,7 +1250,6 @@ def set_batch_size(self, batch_size) -> None:
         super().set_batch_size(batch_size=batch_size)
         self.v = self.rest * torch.ones(batch_size, *self.shape, device=self.v.device)
         self.u = self.b * self.v
-        self.refrac_count = torch.zeros_like(self.v, device=self.refrac_count.device)
```
Hananel-Hazan: I'm not sure why this line has been added

bindsnet/network/nodes.py
```diff
@@ -1218,7 +1218,7 @@ def forward(self, x: torch.Tensor) -> None:
 
         # Add inter-columnar input.
         if self.s.any():
-            x += self.S[:, self.s].sum(dim=1)
+            x += self.S[:, self.s.view(-1)].sum(dim=1)
```
djsaunde: I don't think this will work if you use minibatching. Spikes will have shape `[batch_size, ...]`; I'm guessing you're using `batch_size = 1` in you use case. To get it to work with arbitrary minibatch size, you should replace it with:

```
torch.cat([S[:, s[i]].sum(dim=1)[None] for i in range(s.shape[0])], dim=0)
```

Or a vectorized equivalent of that.

PR #309, opened 2019-08-28
Added ALOV300+ dataset

coopersigrist:
The network added is an ANN, but this is just for benchmarking and working on the conversion task. The paper I am replicating with the network used imagenet challenge set for testing. I implemented this but I want to rework it before I PR.


djsaunde:
@coopersigrist looks like there are test failures. Could you fix them before we can merge?

bindsnet/datasets/alov300.py
```diff
@@ -0,0 +1,318 @@
...
+        envs = os.listdir(target_dir)
+        num_anno = 0
+        print("Parsing ALOV dataset...")
+        for env in envs:
+            env_videos = os.listdir(root + env)
+            for vid in env_videos:
+                if vid in self.exclude:
+                    continue
+                vid_src = self.frame_path + env + "/" + vid
+                vid_ann = self.box_path + env + "/" + vid + ".ann"
```
ELanning: Consider using an f-string instead, as-is this is creating 4 different strings. (`vid + ".ann"` creates a new one which is then added to `"/"` which creates a new one, and so forth)

I bet most interpreters can probably optimize it, but figure I'd mention it.

bindsnet/datasets/alov300.py
```diff
@@ -0,0 +1,318 @@
...
+                frames = os.listdir(vid_src)
+                frames.sort()
+                frames = [vid_src + "/" + frame for frame in frames]
+                f = open(vid_ann, "r")
```
ELanning: Consider using the `with` keyword to open the file, or was it done this way to reduce indentation?

bindsnet/datasets/alov300.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+from PIL import Image
+from glob import glob
+from tqdm import tqdm
+from collections import defaultdict
+from typing import Optional, Tuple, List, Iterable
+from urllib.request import urlretrieve
+from torchvision import transforms
+import xml.etree.ElementTree as ET
```
ELanning: Is this `import` used anywhere? I tried searching but couldn't find where it's used.

Glancing through, a lot of these `import`s don't seem to be used? eg `defaultdict`, the typings, `tqdm`, etc

bindsnet/datasets/alov300.py
```diff
@@ -0,0 +1,318 @@
...
+from torch.utils.data import Dataset
+
+warnings.filterwarnings("ignore")
+
+
+class ALOV300(torch.utils.data.Dataset):
```
ELanning: Is there a reason for using the whole namespace when it's imported as `Dataset` above?

bindsnet/datasets/alov300.py
```diff
@@ -0,0 +1,318 @@
...
+        sample = {"image": curr, "bb": currbb}
+        return sample
+
+    def get_bb(self, ann):
+        """
+        Parses ALOV annotation and returns bounding box in the format:
+        [left, upper, width, height]
+        """
+        ann = ann.strip().split(" ")
+        left = min(float(ann[1]), float(ann[3]), float(ann[5]), float(ann[7]))
```
ELanning: Instead of calling `float` on all these tokens separately, is there a reason the `ann` initialization cannot be this?
```
# Renamed ann to coordinates because it's usually not best practice to re-assign an argument.
coordinates = map(lambda x: float(x), ann.strip().split(" "))
```

bindsnet/datasets/alov300.py
```diff
@@ -0,0 +1,318 @@
...
+        :param download: Specify whether to download the dataset if it is not present
+        :param num_samples: Number of samples to pass to the batch
+        """
+        super(ALOV300, self).__init__()
+
+        # Makes a unique path for a given instance of davis
+        self.root = root
+        self.download = download
+        self.num_samples = num_samples
+        self.frame_zip_path = os.path.join(self.root, "frame.zip")
```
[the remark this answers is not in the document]
coopersigrist: I originally had changed root after I set them, but I removed that so I'll move these too thanks!

bindsnet/datasets/alov300.py
```diff
@@ -0,0 +1,318 @@
...
+class ALOV300(torch.utils.data.Dataset):
+    DATASET_WEB = "http://alov300pp.joomlafree.it/dataset-resources.html"
+    VOID_LABEL = 255
+
+    def __init__(
+        self,
+        root,
+        transform,
+        input_size,
+        codalab=False,
```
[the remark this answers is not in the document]
coopersigrist: Copied over from another dataset, my bad!

bindsnet/datasets/alov300.py
coopersigrist: good catch! Yeah I much prefer f string, surprised I wrote it this way

bindsnet/datasets/alov300.py
coopersigrist: Yeah these were used before refactoring. I may still use TQDM but I'll remove it for now

bindsnet/datasets/alov300.py
coopersigrist: haha, no there is not

bindsnet/datasets/alov300.py
coopersigrist: That is a much better way to do this thanks!

PR #310, opened 2019-09-03
Migrate monitor to a list instead of fixed tensor

k-chaney:
The monitor was originally setup on top of a fixed ```torch.Tensor``` and called cat with every iteration. This turned out to be a large amount of compute. Instead, using a list and just copying the small amount of data turns out to be more speed efficient.


djsaunde:
LGTM

Hananel-Hazan:
Thanks @k-chaney

PR #311, opened 2019-09-16
Update random network baseline to use EnvironmentPipeline

ELanning:


examples/space_invaders/et_space_invaders.py
```diff
@@ -11,7 +11,6 @@
 from bindsnet.network.topology import Connection
 from bindsnet.pipeline import EnvironmentPipeline
 from bindsnet.pipeline.action import select_multinomial
-from bindsnet.analysis.plotting import plot_weights
```
ELanning: Saw this import was not used so I deleted it while I was in the area.

ELanning:
The plotting didn't seem to be working, but I think I'd rather fix that in another PR.

examples/asteroids/random_network_baseline.py
```diff
@@ -120,9 +120,11 @@
...
 try:
     while i < n:
-        pipeline.step()
+        result = pipeline.env_step()
+        pipeline.step(result)
 
-        if pipeline.done:
+        isDone = result[2]
+        if isDone:
             pipeline.reset_()
```
ELanning: It's usually bad practice to call pseudo-private methods directly, but this was already here 🤷‍♂

examples/breakout/breakout_stdp.py
```diff
@@ -55,30 +53,28 @@
 )
 
 
-# Train agent for 100 episodes.
+def run_pipeline(pipeline, episode_count):
```
ELanning: This should probably be written to take advantage of the `EnvironmentPipeline.train()` method instead, but I don't particularly feel like doing that atm (just want to fix deprecated `Pipeline` imports rather than make the code perfect).

Hananel-Hazan:
@ELanning Thank you for your efforts with helping us to update the BindsNET examples.
I will try to add some scripts this weekend. 
Anyhow, ping me when this PR is ready to merge.

sharath:
@ELanning thanks for the PR. We're working on re-organizing / fixing all the examples. The issue for that is [here](https://github.com/Hananel-Hazan/bindsnet/issues/312).

For this PR, could you add a script `breakout/random_network_baseline.py` under breakout (changing the environment to `BreakoutDeterministic-v4`) and then remove the asteroids script/directory?

ELanning:
@sharath done

examples/asteroids/random_network_baseline.py
djsaunde: Change `isDone` to `is_done` (PEP8). Also, AFAIK, methods pre-fixed with underscores pseudo-private, not post-fixed.

ELanning: Good catch, that was changed in another PR.

>Also, AFAIK, methods pre-fixed (not post-fixed) with underscores are pseudo-private.

I believe it's project dependent, unless there's some sort of PEP8 standard for it.
If it's not pseudo-private, the question becomes: why is it `reset_` and not simply `reset`?

djsaunde: Good question. It's because I wanted to have a common function name for resetting `Network`s, `Nodes`, and `Connection`s, and many `Nodes` already had a `reset` attribute (referring to post spike reset voltage) at the time.

ELanning: Suppose originally `Nodes` had a typo and instead of `reset_`, it was called `resat`, and that you had already released this code and people were relying on `resat`.

I would advocate adding a new method called `reset`, keeping the `resat` around and marking it deprecated with a note to use `reset` for new code, and then naming all future reset methods as `reset`, _instead_ of doubling down on `resat` and naming all future reset methods as `resat`.

This is just my opinion, and I renamed `reset_` to `resat` to better illustrate the weirdness of doubling down on an inconsistent name, but I suppose it is what it is.

djsaunde: I don't really see it as "doubling down", more of a lazy design choice when the code was just a grad school project. I don't mind changing the name of the method, but of course `resat` is no good. Maybe `reset_state_variables`?

ELanning: `reset_state_variables` might be good, depends how general/closed the method should be.

Some classes might want to do other things then just resetting the variables, but I suppose they could just have two different reset methods. 🤷‍♂

djsaunde:
@ELanning @Hananel-Hazan Is this all set to merge?

Hananel-Hazan:
I'm ready to merge, waiting for ELanning acknowledge

ELanning:
Roger that, all systems ready to go.

PR #324, opened 2019-10-10
Renaming vars, standardizing doc strings, test Python 3.7 on Travis CI, etc.

djsaunde:
Note: there are some backwards compatibility-breaking changes.


Hananel-Hazan:
Nice documentation improvement @djsaunde

bindsnet/conversion/conversion.py
```diff
@@ -82,8 +82,8 @@ def forward(self, x: torch.Tensor) -> Dict[nn.Module, torch.Tensor]:
 class SubtractiveResetIFNodes(nodes.Nodes):
     # language=rst
     """
-    Layer of `integrate-and-fire (IF) neurons <http://neuronaldynamics.epfl.ch/online/Ch1.S3.html>`_ using reset by
-    subtraction.
+    Layer of `integrate-and-fire (IF) neurons
+    <http://neuronaldynamics.epfl.ch/online/Ch1.S3.html>` using reset by subtraction.
```
sharath: the link is broken

bindsnet/conversion/conversion.py
```diff
@@ -197,7 +197,8 @@ def set_batch_size(self, batch_size) -> None:
 class PassThroughNodes(nodes.Nodes):
     # language=rst
     """
-    Layer of `integrate-and-fire (IF) neurons <http://neuronaldynamics.epfl.ch/online/Ch1.S3.html>`_ with using reset by
+    Layer of `integrate-and-fire (IF) neurons
+    <http://neuronaldynamics.epfl.ch/online/Ch1.S3.html>`_ with using reset by
     subtraction.
```
sharath: is this accurate?

dee0512: No, it is not. Why not move these nodes to nodes.py?

ELanning: >Also, why not move these nodes to nodes.py?

I imagine it depends on the intent of the author. If it is a general node that lots of 3rd party consumers might want to use, it makes sense to put it in the `nodes` module. If it is a one off node that is specific to the task at hand, and does not have much utility to consumers, it may be better to keep it local to this file and _not_ share it as part of the public API, to keep the public API tight and simple.

dee0512: I think the subtractive IF neurons is a general node

bindsnet/conversion/conversion.py
djsaunde: It works for me. Maybe the website was down temporarily?

bindsnet/conversion/conversion.py
djsaunde: I think the subtractive IF neurons are specific to the `conversion` module, actually. But yeah, the doc string is wrong.

PR #326, opened 2019-10-20
fixed docs in netwrok.run()

dizcza:
`torch.flatten()` was added in v0.4.1. I changed this because I experienced some error with `.view(-1)` but now I cannot reproduce it, so it can be neglected if you wish.
The main fix is in `network.run()` docs and typo in `inpts`.

Update.
Also, I forgot to add a line in `network.run()` that `time` axis is obtained from `encoding` module, cause it was not obvious for me.

### DIFF
--- a/bindsnet/encoding/encodings.py
+++ b/bindsnet/encoding/encodings.py
@@ -68,7 +68,7 @@ def bernoulli(
     assert (datum >= 0).all(), "Inputs must be non-negative"
 
     shape, size = datum.shape, datum.numel()
-    datum = datum.view(-1)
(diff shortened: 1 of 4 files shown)


ELanning:
>torch.flatten() was added in v0.4.1. I changed this because I experienced some error with .view(-1) but now I cannot reproduce it, so it can be neglected if you wish.

`flatten` does read a bit more idiomatic. Nice changes LGTM 👍

Hananel-Hazan:
Thanks @dizcza for helping us to improve BindsNET

PR #327, opened 2019-10-21
input bug in pipeline

Jbwasse2:
While trying to run bindsnet/examples/breakout/breakout.py, I would get error involving inputs in the given file. After applying this fix, breakout runs again.


Hananel-Hazan:
Thanks! good catch

Issue #333, opened 2019-11-02
Does bindsnet support to forward two different input sets simultaneously ?


ati6868:
Hi
As i said in my previous issue, i created 3 layers(one layer of `Input` nodes and 2 layers of `LIFNodes`)
Connections between layers are like this :  A------>B<=======>C
but now i am going to feed mnist images to layer(A) that will be forwarded to layer(B) , on the other hand i'm going to feed my pre-defined input spike trains to layer(B) [coming from layer(C)].
then i will compare mnist images poisson spike trains with new generated spike trains in layer(B)

but when i'm going to run this command:
`network.run(inpts={"C": datum}, time=time, input_time_dim=1)`
it gives me this error:
`File "C:\python\Python37\lib\site-packages\bindsnet\network\network.py", line 344, in run`
`self.layers[l].forward(x=inpts[l][t])
KeyError: 'A'`

that's true that i created feedback connections but it only works for inputs coming from first layer(A) , but not for starting from layer(C) by its own inputs.
I would be grateful if give me an idea on this

Issue #332, opened 2019-10-30
how to create feedback connections in bindsnet?


ati6868:
Hi
I'm trying to create a network of two layers while the network has feedforward and feedback connections
`feedback_connection = Connection(
    source=target_layer,
    target=source_layer,
    w=0.05 + 0.1 * torch.randn(target_layer.n, source_layer.n),
)`

but it shows me this error:
`    network.run(inpts=inpts, time=time)
  File "C:\python\Python37\lib\site-packages\bindsnet\network\network.py", line 383, in run
    inpts.update(self._get_inputs())
  File "C:\python\Python37\lib\site-packages\bindsnet\network\network.py", line 230, in _get_inputs
    inpts[c[1]] += self.connections[c].compute(source.s)
  File "C:\python\Python37\lib\site-packages\bindsnet\network\topology.py", line 188, in compute
    post = s.float().view(s.size(0), -1) @ self.w + self.b
RuntimeError: size mismatch, m1: [100 x 1], m2: [100 x 1000] at C:\w\1\s\windows\pytorch\aten\src\TH/generic/THTensorMath.cpp:197`

![bindsnet](https://user-images.githubusercontent.com/52522391/67859332-9c40b880-fb30-11e9-92ed-4a0beeff37ea.JPG)

i would be grateful if someone could help me on this

djsaunde:
Hi @ati6868, at the moment, it makes no sense to connect to a layer of `Input` nodes, they simply discard their input. `Input` nodes spikes are set by the user by passing them to `network.run(inputs=...)`.

You could instead create a network with a layer of `Input` nodes (A), connected to a layer of `LIFNodes` (B), which is recurrently connected to itself, or connected to a third layer of `LIFNodes` (C) which connects back to B.

It might be good to support adding support for your use case, though; I'll look into it.

ati6868:
thank you @djsaunde  for your response . my model is supposed to use predictive coding that's why i need to use feedback connections for propagating predictions of lower layers generated by upper layers . 
if it's not possible , Could i have two layers of `LIFNodes` instead of " one layer of `Input` and one layer of `LIFNodes` " ?

ati6868:
thank you very much @djsaunde 
just how can i be informed when u do that because i'm working on my project with bindsnet??

ati6868:
By the way ....As you said  that layer of `Input` nodes works based on user-specified spiking behavior.
[...]

PR #335, opened 2019-11-04
Refactoring Input nodes to support incoming connections; removing RealInput nodes (subsumed by Input nodes); adding support for input to arbitrary layers.

djsaunde:
This PR is a little tricky, so I don't expect it to get merged right away. Addresses use cases discussed in #332 and #333. Namely, this adds support for:

- Incoming connections to `Input` nodes that do something. `Input` nodes now use input from incoming connections instead of discarding it in favor of what's passed by `network.run(inputs=...)`. NB: users have to be a bit more careful with data types in this situation.
- The output of `Input` nodes can now have any type; this subsumes the `RealInput` nodes class, which has been removed as a result.
- Any kind of node can take input passed to `network.run(inputs=...)`. For example, you can create a network with only `LIFNodes` (no `Input` nodes), and inject input directly to them.


ati6868:
@djsaunde , thanks for your big help, that's perfect 
Well done:)

PR #355, opened 2020-03-03
Fix bugs to make CNN converted to SNN successfully

WeihaoTan:
modified:   bindsnet/conversion/conversion.py	
modified:   bindsnet/network/network.py
modified:   bindsnet/network/topology.py

Because I only run the CNN converted to SNN code for test. The change may be incompatible with other code. More tests are needed. See comments to get the explaination of the revision.


Hananel-Hazan:
Thank you for your PR and the work done. 
I'll check it soon

Issue #362, opened 2020-05-01
supervised MNIST example accuacy and GPU issues


Thonner:
Hi BindsNet
I am doing a master project where I am building a neuromorphic processor. I'm am (trying) to use your platform to train a network using Bindsnet, as this is the most promising SNN simulator that stays purely in the spiking domain. Long story to say that I have been playing around supervised MNIST examples, as I would like to use this to test my hardware implementation. But I'm seeing some issues.

If you want to use GPU it crashes in the `poisson` function in `encoding.py`. where it expects `cuda:0` device but gets `cpu`. Not sure but I think its the dataloader that needs to be sent to the right device.

Next: The accuracy measurements that I provided in the training loop can not be right. I have tested this the following way. First i train a network over 5000 inputs, using the provided training loop. This loop reports 100% accuracy after about 600 inputs. After training, I turn off training in the network. Then i test the accuacy after training with the following loop:
```python
pbar = tqdm(enumerate(dataloader))
    for (i, datum) in pbar:
        if i > 1000:
            break
        image = datum["encoded_image"]
        label = datum["label"]
        inputs = {"X": image.view(time, 1, 1, 28, 28)}
        network.run(inputs=inputs, time=time)

        out_spikes = network.monitors["Ae_spikes"].get("s")
        class_spike=torch.zeros(10)
        for i in range(time):
            for j in range(n_neurons):
                if out_spikes[i,0,j]:
                    class_spike[j//10] += 1
        
        aVal, maxInd = class_spike.max(0)
        if maxInd == label[0]:
            hit += 1

    
    acc = hit/1000
    print("\n accuracy: " + str(acc) +"\n")
```
This test returns accuracy around 65% depending on the seed. So what is the accuracy that the training loop is stating? To me, my loop should give the actual accuracy.

Sorry for the long issue! Your platform seems really cool! Therefore it is also a shame that your display piece is not running properly. I hope that you'll take the time to have a look at my complaints
- Anthon

Hananel-Hazan:
Thank you for bringing this up. You are right, there was a couple of issues with the example that prevent it to work under GPU. 
The data loader has been fixed. 

The accuracy reported in the example is for the training process. It shows that the network reacts as expected. Once the training completes the supervise mechanism that clamps the right neuron and forces it to fire no longer valid. 
[...]

PR #364, opened 2020-05-05
GPU device miss-match in encoder fix

Thonner:
After changes made from issue https://github.com/BindsNET/bindsnet/issues/362 the Poisson encoder still crashed due to device miss-match. The change in this pull request is directly taken from your reservoir.py MNIST example. 
That said - it actually runs slower on my machine on GPU. Granted I have had a hard time setting up cuda on my ubuntu.


Hananel-Hazan:
Thanks for the PR!

Some examples in BindsNET will run slower on GPU due to calculation overhead. When the size of the network is small and there is more data to be swap around from CPU to GPU there is no benefit of using GPU. 
GPU will shine when the Network is large and the data stored mostly in the GPU, then the calculation overhead will be negligible compare with CPU calculation. That is true to any GPU calculation.

PR #369, opened 2020-05-12
Add summary() function for the Network class.

SimonInParis:
Summarizes informations about a Network in text format.


Hananel-Hazan:
Thanks @SimonInParis

PR #370, opened 2020-05-15
update 20200515 GPU-plot

thomasaimondy:


thomasaimondy:
update others gpu and plot functions

thomasaimondy:
update others' plot function

PR #384, opened 2020-06-17
faster LIF forward

SimonInParis:
Hell yeah.


Hananel-Hazan:
Thank you @SimonInParis for this great work! this is quite a boost to the performance!

PR #388, opened 2020-06-27
replace a kind of variable `device_id` to `device`

DenDen047:
I couldn't run the sample program.
I think it should be `device` instead of `device_id`.


Hananel-Hazan:
Thank you for catching this and submit PR!

PR #415, opened 2020-09-23
Indices exchange in plot_spikes function

ruteee:
Indices exchange in plot_spikes function, so that time varies from 0 to time param, and n_neurons vary from 0 to the number of neurons at each layer.

### DIFF
--- a/bindsnet/analysis/plotting.py
+++ b/bindsnet/analysis/plotting.py
@@ -98,13 +98,13 @@ def plot_spikes(
     if time is None:
         # Set it for entire duration
         for key in spikes.keys():
-            time = (0, spikes[key].shape[0])
+            time = (0, spikes[key].shape[1])
             break
 
     # Use all neurons if no argument provided.
     for key, val in spikes.items():
         if key not in n_neurons.keys():
-            n_neurons[key] = (0, val.shape[1])
+            n_neurons[key] = (0, val.shape[0])
 
     if ims is None:
         fig, axes = plt.subplots(n_subplots, 1, figsize=figsize)


Hananel-Hazan:
Grate Thank you

davidAmiron:
When I run the example code at the end of [this tutorial](https://bindsnet-docs.readthedocs.io/guide/guide_part_i.html) I get incorrect output; the number of neurons and time are switched, it seems to be as a result of this PR. When I reverse what was done here, I get the correct output. I see this is related to the reservoir mnist example.

Perhaps this is a problem in data formatting in the reservoir mnist example, or maybe some style changed and the example in that tutorial must be updated, I am not sure.

Hananel-Hazan:
Thanks @davidAmiron for pointing it out, you are right, the reservoir example had wrong plotting dimensions. Fixed in  #417 

Regarding the code in the tutorial, its seems running fine
image

ruteee:
Hi, David. 

I did the PR for this update. I tested the code example of the tutorial you mentioned and you are right the indices seem to be switched. 

However if run reservoir example with indices switched back to their original positions (like the old version), the plot_spikes and the plot_voltages generate the incorrect output. That is with the neurons in the place of time, and vice-versa.

I agree there is probably something happening in data formatting in the reservoir example. I apologize for the mistake and suggest the reversion of PR.

ruteee:
@Hananel-Hazan should I do the PR for undoing the modification?

davidAmiron:
I have a branch with the changes in it if you would like me to submit it

Hananel-Hazan:
@davidAmiron can you send screenshot of the plot issue you having?

davidAmiron:
Running the example from the tutorial on master, I have the following:
image
For the ```plot_spike function```, the number of neurons and number of timesteps are switched. If you run the example and print out ```spikes[key].shape``` right before line 101 of [plotting.py](https://github.com/BindsNET/bindsnet/blob/master/bindsnet/analysis/plotting.py) and ```val.shape``` before line 107 of the same file, you can see that the order of number of neurons and number of timesteps is switched from what is being read in those lines, respectively.

Hananel-Hazan:
I am sorry, I dont get the same plotting. Are you sure you are using the latest update? The screenshots that I getting from the tutorial look fine, they are posted here.

davidAmiron:
Hmm. I am working directly off of master. If you are using a previous pip installation maybe #415 isn't on your machine, because it is only 12 days old?

Hananel-Hazan:
Sorry, something went wrong with my GitHub merge that mask this update. @ruteee please revert the change. Thank you @davidAmiron for pointing it out

ruteee:
yes, for sure.

PR #429, opened 2020-11-16
Speedups3

SimonInParis:
let's see


Hananel-Hazan:
Thank you Simon for the performance boost and cleaning up the code!

PR #430, opened 2020-11-23
Update action.py

dartl0l:
Function returns tensor instead of integer, so it fails in EnvironmentPipeline.


Hananel-Hazan:
Thank you @dartl0l, check the PR and will let you know

Hananel-Hazan:
Thank you for the PR

PR #443, opened 2020-12-30
Implement Cumulative Spike Response Model Nodes

sidharthanup:
A vectorized implementation of the Spike Response Model (Gerstner) that models the combined effects of input current and the spike after potentials of several previous spikes, over a time window, using response and refractory kernels. This model is being used in many of the current SNN supervised learning algorithms like [SLAYER](https://github.com/bamsumit/slayerPytorch) and [TSSLB](https://github.com/stonezwr/TSSL-BP). I will be sending follow up PRs to implement those learning rules once this is merged.
Signed-off-by: Sidharth Anupkrishnan <sanupkri@redhat.com>

### DIFF
--- a/bindsnet/network/network.py
+++ b/bindsnet/network/network.py
@@ -4,7 +4,7 @@ from typing import Dict, Optional, Type, Iterable
 import torch
 
 from .monitors import AbstractMonitor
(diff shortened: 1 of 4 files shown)


Hananel-Hazan:
Great addition @sidharthanup

PR #455, opened 2021-02-14
Fix input plot issue

jundeli:
Fixed the mnist input plot issue that index will go out of limit when labels list is reset to empty.


Hananel-Hazan:
Thank you for the PR, but I am afraid that this is not indexing issue.

Issue #467, opened 2021-03-23
Google Speech Commands and BindsNET


hansemandse:
Hi! I am trying to train a very small feed forward network much alike your example with MNIST with the dataset replaced by Google Speech Commands (GSCD), but I can't seem to make the network learn anything (see `train.py`). My source code is available [here](https://github.com/hansemandse/KWSonSNN/tree/master/netwkdev). 

The network I am using is a scaled version of the `DiehlAndCook2015` network (see `kwsonsnn/model.py`) made to fit on a custom hardware accelerator. It works with MNIST achieving roughly 70% accuracy in a single epoch. I believe my dataset (see `kwsonsnn/dataset.py`) is correctly preprocessed in a style similar to your implementation of Spoken MNIST. The regular PyTorch model included in the repo (see `trainpt.py`) achieves 90% accuracy on GSCD indicating that data is not to blame.

I have tried multiple different parameter configurations in terms of weight normalization and learning rates, but they do not seem to improve the network's performance.

So, am I doing something wrong? And do you by any chance have some benchmark results for either Spoken MNIST or GSCD produced with BindsNET?

System specifications:
- Python3 3.7.6
- BindsNET 0.2.7
- numpy 1.18.1
- PyTorch 1.6.0+cpu

Hananel-Hazan:
Thank you for using BindsNET, I think this is more of a discussion and not an issue in BindsNET. 

> am I doing something wrong?

Its hard to say, tuning a network parameters (any network SNN or ANN) is not an easy task and require a lot of trial and error from your part. I can only try to help. I will start by upgrading BindsNET to 0.2.8 (instruction in the main repo page).
Then, please add some plots: weights, voltage curves of the neurons, spikes of #neuron vs time, and input reconstruction. Hopefully the plots will show you how to approach the parameter tuning task.

A brief glimpse on your code, I think that your normalization value is too high. Note that the normalization value choosen by `DiehlAndCook2015` was 10% of the input neurons, that also reflect the B/W pixel ratio in MNIST.

 I couldn't spot an error 


>  do you by any chance have some benchmark results for either Spoken MNIST or GSCD produced with BindsNET?

Sadly, no, but If you run them, I will be happy to add them to the repo with your name as a contributor.

hansemandse:
Thanks for replying so quickly!

[...]

PR #474, opened 2021-04-14
fix yticks in plot_spikes

petermarathas:
This adds two y axis labels to the scatter plots in `plot_spikes`, corresponding to the beginning and end of the n_neuron range. It should fix an issue uncovered in https://github.com/BindsNET/bindsnet/issues/467 that `plot_spikes`was not showing spikes for intermediate layers. The issue was that the y axis would have a small range around the spikes of the initial data entry, and this range would not cover the full span of possible neurons.

### Before
Screen Shot 2021-04-14 at 12 02 06 AM

### After
Screen Shot 2021-04-14 at 12 01 31 AM


Hananel-Hazan:
Thanks @petermarathas

PR #489, opened 2021-05-31
Update random_network_baseline.py

het-25:
The output layer should have 4 neurone: one for each possible action on the joystick.


Hananel-Hazan:
Good catch!

PR #492, opened 2021-06-10
Update reservoir.py

cearlUmass:
Added additional comments for clarification, and fixed dimensions for monitors during the testing dataset (lines 254 & 259)


Hananel-Hazan:
Good catch! thanks @cearlUmass

Issue #496, opened 2021-06-23
Set torch version to 1.8.1


GutiPerez:
Hello

_torch_ has been updated to version 1.9.0, and as the requirement for _torch_ is >=1.5.1, the package is updated when installing BindsNET. Because some changes made to _torch_ in this new version, there are errors because some imports does not exist anymore, like _container_abcs_ and _int_classes_ from _torch._six_ in the file _collate.py_ (although it occurs also in another files).

I guess that you should fix _torch_ version to 1.8.1 until the impact from this new version is evaluated in more detail.

Hananel-Hazan:
Good catch! thank you for lets us know. 
push freeze 1.8.1 #497

Hananel-Hazan:
Fix the issues cause by PyTorch 1.9.0 #499 
Let me know if you have any issues

GutiPerez:
There are still some issues. 

In the file _collate.py_, on line 76, in the function _isinstance_, the argument _int_classes_ does not exists, as you have deleted the import. But you forgot to change that argument. Putting a simple "_int_" instead solves the issue, as the import from torch._six was importing that.

Then, in the same file, but now on lines 80 and 84, in the function _isinstance_ again, the argument _container_abcs_ does not exist, again, because the import was deleted. You imported _collections_, but forgot to change those arguments. Make the following import fixes the issue: _"import collections.abc as container_abcs"_.

That is only on the file _collate.py_, but there could be more instances of the same problems in another files.

Hananel-Hazan:
Good catch and thanks for the help! 
#500

PR #500, opened 2021-06-27
Fix issues cause by torch._six upgrade

Hananel-Hazan:


Hananel-Hazan:
Thanks for Miguel Gutierrez Perez #496

PR #513, opened 2021-08-28
Switch to Poetry

danielgafni:
- Switch to [Poetry](https://python-poetry.org/) dependencies manager
 - Add docker image GPU support (FROM nvidia/cuda:11.1-base)
 - Add optional "production" docker build without development dependencies (controllable with the DEPS=[development, production] build arg)
 - Add optional docker build with a fix for nvidia 30xx GPUs (controllable with the NVIDIA_30XX=[true, false] build arg)
 - Local Black pre-commit hook (benefits - fixed black version)
 - CI fixes
 - Required Python version changed 3.6 -> 3.7 (because of scikit-image incompatibility):

```
The current project's Python requirement (3.6.13) is not compatible with some of the required packages Python requirement:
    - scikit-image requires Python >=3.7, so it will not be satisfied for Python 3.6.13
  
  Because scikit-image (0.18.3) requires Python >=3.7
   and no versions of scikit-image match >0.18.3,<0.19.0, scikit-image is forbidden.
  So, because bindsnet depends on scikit-image (^0.18.3), version solving failed.
```


danielgafni:
A few python files were changed but that's just Black formatting

danielgafni:
@Hananel-Hazan should I also update `CONTRIBUTING.md` with poetry related instructions?

Hananel-Hazan:
@danielgafni , Please do that

Hananel-Hazan:
Thanks @danielgafni for the great work!

PR #532, opened 2021-12-29
docs: Fix definition of wmax

amirHossein-Ebrahimi:
- Update doc section to be more accurate about argument wmax


Hananel-Hazan:
Good catch, thank you for helping with the documentations!

PR #534, opened 2022-01-01
Fix WeightDependentPostPre Post-synaptic update

ArefAz:
The difference of `wmax` and the actual weights had been mistakenly calculated as `(self.wmax - self.connection.wmin)`, which was a fixed value.


Hananel-Hazan:
Good catch! thank you for helping out!

ArefAz:
I just noticed the same error in the `_conv1d_connection_update` method, line `483`.

Thank you!

Hananel-Hazan:
Right! just pushed fix for that too. 
Thanks again

PR #536, opened 2022-01-18
Localconnections

hafezgh:
Add three local connection classes (1D, 2D, and 3D) supporting multi-channel inputs alongside MNIST example files.


hafezgh:
Fixed the MaxPool batch size issue

PR #540, opened 2022-02-07
Fix calling to NoneType object

### LINKED ISSUE #539: Call NoneType object!
Here is a part of `environment_pipeline.py`:
```
# Place the observations into the inputs.
if self.encoding is None:
    obs = obs.unsqueeze(0).unsqueeze(0)
    obs_shape = torch.tensor([1] * len(obs.shape[1:]), device=self.device)
    inputs = {
        k: self.encoding(
            obs.repeat(self.time, *obs_shape).to(self.device),
            device=self.device,
        )
        for k in self.inputs
    }
else:
    obs = obs.unsqueeze(0)
    inputs = {
        k: self.encoding(obs, self.time, device=self.device)
        for k in self.inputs
    }
```
As you can see, when `self.encoding` is None, `self.encoding` is called at line 6.

navidkashi:
Fixes #539 issue by removing the unnecessary calls to `self.encoding`.


Hananel-Hazan:
Thanks @navidkashi

PR #543, opened 2022-03-14
Update dot_simulator.py

williamyao66:
File naming error.


Hananel-Hazan:
Thanks @williamyao66 good catch!

PR #552, opened 2022-05-17
Vectorial learning rates for all UpdateRules.

ignaciogavier:
:param nu: Single or pair of learning rates for pre- and post-synaptic events. It also accepts a pair of tensors to individualize learning rates of each neuron. In this case, their shape should be the same size as the connection weights.


Hananel-Hazan:
Thanks @ignaciogavier good catch!

PR #568, opened 2022-08-01
Stdp nids

dovahkiin1995:
STDP code added for UNSW dataset.
in examples/nids


Hananel-Hazan:
Please send me an email to Hananel@hazan.org.il and we can talk about this PR.

PR #606, opened 2022-12-20
Remove unused argument `input_time_dim`

0xnurl:
`input_time_dim=1` is used througout the examples but never eventually used (error never occurs because it is eventually absorbed in some `kwargs`). Removing it to avoid confusion.


Hananel-Hazan:
Thank @0xnurl for the PR! always fun to clean the code from old leftovers.

PR #660, opened 2023-10-12
Update reservoir.py

AXYZdong:
- Added additional comments for plotting in train progress (lines 147, 155, 161, 167)
- Fixed a minor bug in the source code at line 239 (see line 244 of the updated code). _Don't need to calculate the sum of each column of the array, i.e. don't need to use the `.sum(0)` ._


Hananel-Hazan:
Good catch

PR #678, opened 2024-03-29
Fixed collections issue

bmosk54:


Hananel-Hazan (changes requested):
please revert the changes and change line 10 to
`import collections.abc`

in line 77 add abc between collections to Mapping
`collections.abc.Mapping`

Hananel-Hazan:
Fantastic, thank you for fix!

PR #712, opened 2025-07-31
docs: Update README badges - Collaboration Network

neural-loop:
I adjusted it a bit so that it is the same size as the other badges in the README.md

image

Thank you, and I look forward to future collaborations!


Hananel-Hazan:
Thanks @neural-loop

PR #743, opened 2026-03-27
fix: correct bernoulli_loader kwarg, remove duplicate import, fix mutable default

haoyu-haoyu:
## Summary

- **Fix `bernoulli_loader` reading wrong kwarg** in `bindsnet/encoding/loaders.py`: `kwargs.get("dt", 1.0)` should be `kwargs.get("max_prob", 1.0)`. The docstring documents `max_prob` as the parameter for maximum spike probability per Bernoulli trial, but the code was reading `dt` instead.

- **Remove duplicate `import numpy as np`** in `bindsnet/network/monitors.py`.

- **Fix mutable default argument** `pipeline: list = []` → `pipeline: Optional[list] = None` in `MulticompartmentConnection.__init__` (`bindsnet/network/topology.py`). Mutable defaults are shared across all instances.

## Test plan

- [ ] `bernoulli_loader` now correctly reads `max_prob` parameter
- [ ] No behavioral change for existing code that doesn't pass `max_prob` (default remains 1.0)
- [ ] Existing tests should pass without modification

### DIFF
--- a/bindsnet/encoding/loaders.py
+++ b/bindsnet/encoding/loaders.py
@@ -26,7 +26,7 @@ def bernoulli_loader(
     :param float max_prob: Maximum probability of spike per Bernoulli trial.
     """
     # Setting kwargs.
-    max_prob = kwargs.get("dt", 1.0)
+    max_prob = kwargs.get("max_prob", 1.0)
 
     for i in range(len(data)):
         # Encode datum as Bernoulli spike trains.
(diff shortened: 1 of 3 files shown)


Hananel-Hazan:
The `bernoulli_loader` `max_prob` fix here is included in the PR referenced above (commit `593d8dce`), which is on its way to `master`. Closing as resolved once that merges — thanks @haoyu-haoyu for the catch.

PR #755, opened 2026-05-06
Fix eth_mnist example

andreapignaz:
I’m new to the project and noticed that the eth_mnist example does not work. This one-line change fixed it for me.
The original version tried to access weights from a MulticompartmentConnection in a way is not (anymore?) supported.


Hananel-Hazan:
Great catch! thanks you

